PR #109, opened 2024-09-30
Fix Earthile release target in Beluga vs Nav2 benchmark

marcoshuck:
### Proposed changes

Running `./tools/earthly ./src/benchmarks/beluga_vs_nav2+release` doesn't work:

```
Error: src/benchmarks/beluga_vs_nav2/Earthfile:75:4 parse flag args: invalid argument COPY
in        github.com/Ekumen-OS/lambkin/src/benchmarks/beluga_vs_nav2:main+release --distro= --rosdistro=
```

This can be fixed by removing the backslash from the previous command, suggestion by @hidmic 

#### Type of change

### Checklist

### Additional comments

### DIFF
--- a/src/benchmarks/beluga_vs_nav2/Earthfile
+++ b/src/benchmarks/beluga_vs_nav2/Earthfile
@@ -72,7 +72,7 @@ build:
 release:
     ARG distro=jammy
     ARG rosdistro  # forward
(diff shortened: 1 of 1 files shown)


hidmic (approved):
Good catch! Ready to go after signing the DCO.

PR #116, opened 2024-10-31
Add package to evaluate the impact of the beam skipping feature in nav2

glpuga:
### Proposed changes

Crude benchmark package to compare nav2-vs-nav2, likelihood vs likelihood_prob vs beam_skip.

This is part of a fact finding campaing for https://github.com/Ekumen-OS/beluga/issues/187 .

#### Type of change

### Checklist

### Additional comments


src/benchmarks/beam_skipping_evaluation/reports/nominal_report/sections/platform.inc
```diff
@@ -0,0 +1,129 @@
...
+
+.. list-table:: Real robot datasets
+   :widths: 25 70
+   :header-rows: 1
+
+   * - Name
+     - Description
+   * - `Willow Garage dataset <https://web.archive.org/web/20151207202459id_/http://cgi.cs.duke.edu/~mac/papers/iros12_sm.pdf>`_
+     - Large office environment with multiple recordings along different trajectories and times, including multiple rooms and hallways.
+   * - `TorWIC SLAM dataset <https://github.com/Viky397/TorWICDataset>`_
```
hidmic, line 20: @glpuga do all these datasets apply here?

hidmic (commented):
This is great! Two thoughts. 
- We can but we don't necessarily have to add another package for a different benchmark. We can accumulate benchmarks in a single package too.
- Tied to the above, eventually I would very much like to move Beluga benchmarks to Beluga. Putting all benchmarks in a single package will simplify that transition while keeping the Beluga repo sane. Either that or we have to start considering a separate Beluga benchmarks repository.

glpuga:
> This is great! Two thoughts.
> 
> * We can but we don't necessarily have to add another package for a different benchmark. We can accumulate benchmarks in a single package too.
> * Tied to the above, eventually I would very much like to move Beluga benchmarks to Beluga. Putting all benchmarks in a single package will simplify that transition while keeping the Beluga repo sane. Either that or we have to start considering a separate Beluga benchmarks repository.

This is not necesarily intended to merge. It's just a quick and dirty benchmark draft to share with @DPR00 to get him up to speed to evaluate https://github.com/Ekumen-OS/beluga/issues/187 , to spare him hitting the learning curve head on .

We can later decide what we want to do with this.

PR #130, opened 2026-02-23
Phase 2 Next Gen implementation of simple sequential script

teresa-ortega:
### Proposed changes

I have implemented the initial structure for the Python benchmarking scripts and the ROS 2 launch system. This includes the core logic for running automated tests and the Python launchers to handle parameter overrides (particles, laser range, etc.).

* The user is responsible for defining: the input file paths (reference rosbag, map, and QoS file), as well as the configurations to evaluate (laser models, number of particles, clock rate, and number of iterations).
* Lambkin is a process orchestrator that: automatically manages the execution of the entire benchmark. It launches the Beluga node, plays the rosbag, records the resulting topics, synchronizes the termination of all processes, and runs the evaluation tools (evo_traj, evo_ape) to calculate the metrics.
* Folder structure: lambkin automatically creates a directory structure to save all generated artifacts in an organized way, which looks like this:
```
benchmarking_results/
    └── <sensor_model>_p<num_particles>/
        └── iter_<number>/
            ├── bag/      (Recorded rosbags)
            ├── tum/      (TUM trajectories)
            └── ape/      (Error metrics & zip)
    └── plots_ape/    (Aggregated APE plots across all iterations)
```

#### Type of change

#### How to test

To test these changes, please follow these steps:

Download necessary artifacts: Ensure you have the following files from the internal storage:

1.    Download necessary artifacts:
    Ensure you have the following files from the internal storage:

        * Rosbag: /beluga/hq_simulation/bagfiles/hq_simulation_segment_0

        * Map: /beluga/hq_simulation/artifacts/hq_simulation_segment_0/map.yaml

        * Groundtruth: /beluga/hq_simulation/artifacts/hq_simulation_segment_0/groundtruth.tum

2. Build the Docker image:

```bash
docker compose up -d lambkin_dev
```
3. Start the container:

```bash
docker compose exec -it lambkin_dev bash
```

4. Install dependencies: Inside the container, run:

```bash
rosdep install --from-paths lambkin_ros2 --ignore-src -r -y
```
5. Install Python dependencies:

```bash
uv sync
``` 
 
6. Build the workspace:

```bash
colcon build --symlink-install
source install/setup.bash
```

7. Run the benchmarking script:

```bash
uv run src/lambkin/lambkin.py
```

### Checklist

### Additional comments


src/lambkin/test/test_process_management.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python3
+"""Unit tests for the process management functions in lambkin.py."""
+
+import os
+import subprocess
+import sys
+import time
+import unittest
+
+sys.path.append(os.path.join(os.path.dirname(__file__), "..", "src"))
```
xaru8145: We should avoid sys.path.append in our tests. By properly configuring our `pyproject.toml` and ensuring we have an `__init__.py` in src/lambkin, we can rely on standard imports. This makes our test suite portable and professional.

src/lambkin/src/lambkin.py
```diff
@@ -0,0 +1,240 @@
...
+    for variation in make_variations():
+        for it in range(num_iterations):
+            run_iteration(variation, it)
+            time.sleep(2)
+
+
+if __name__ == "__main__":
+    if len(sys.argv) > 1 and sys.argv[1] == "test":
+        sys.argv.pop()
+        unittest.main()
```
xaru8145: We should remove the manual `unittest` trigger from the `if __name__ == "__main__"`: block. Then, we have `pytest` defined in our `pyproject.toml` and a dedicated `test/` directory (which we already have), we should run tests using the CLI (e.g., uv run pytest) rather than passing a "test" argument to the benchmark script

src/lambkin/src/lambkin.py
```diff
@@ -0,0 +1,240 @@
...
+
+def plot_ape_metrics(ape_path: str):
+    """Plots the Absolute Pose Error (APE) metrics using evo_res.
+
+    Args:
+        ape_path (str): The path to the results zip file generated by evo_ape.
+    """
+    cmd_list = ["evo_res", ape_path]
+    print(f"{cmd_list}\n")
+    subprocess.run(cmd_list)
```
xaru8145: Suggest wrapping this into a `execute_foreground_process` since you also need it in `evo_ape` function.

lambkin_ros2/launch/beluga.launch.py
xaru8145, line 1: You are not launching lambkin, you are launching Beluga. Name is misleading.

lambkin_ros2/launch/beluga.launch.py
```diff
@@ -0,0 +1,70 @@
...
+        package="beluga_amcl",
+        executable="amcl_node",
+        name="beluga_amcl",
+        output="screen",
+        parameters=[
+            {
+                "laser_model_type": LaunchConfiguration("laser_model_type"),
+                "max_particles": LaunchConfiguration("max_particles"),
+            }
+        ],
```
xaru8145, line 56: If only these 2 parameters are set, the rest will take the default value. I would encourage you to define a launch argument for the config path and set it here. I think that if you have the same parameter set via config file and via launch, the launch one takes precedence but you should double check.

src/lambkin/src/lambkin.py
```diff
@@ -0,0 +1,240 @@
...
+num_iterations = 0
+clock_rate = 10
+qos_file_path = "/ws/beluga/beluga_example/bags/qos_override.yaml"
+
+laser_models = ["likelihood", "beam"]
+num_particles = [1, 10, 100, 1000, 10000]
+clock_rate_options = ["--clock-rate", str(clock_rate)]
+qos_options = ["--qos-profile-overrides-path", qos_file_path]
+clock_option = "--clock"
+topics_interested = ["/pose", "/tf", "/tf_static"]
```
xaru8145: I suggest defining the global variables with capital letter so you can distinguish them easily in your code.

xaru8145: topics_interested for what? record, play? Be more explicit to avoid confusion.

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,240 @@
...
+    """Executes a shell command in the background.
+
+    Args:
+        full_cmd_list (list[str]): The command and its arguments as a list of strings.
+        log_file (str, optional): Path to a output log file.
+
+    Returns:
+        subprocess.Popen: The running background process.
+    """
+    print(f"{full_cmd_list}\n")
```
xaru8145: Some functions have a print some others do not. I suggest making the print in all applicable functions and add a `dry_mode` optional argument that defaults to false. If set to true, you just print and do not execute any process. If set to false, no print and you execute the processes. This allows you to quickly test the workflow and commands without waiting for the entire processes orchestration to happen.

src/lambkin/src/lambkin.py
```diff
@@ -0,0 +1,240 @@
...
+
+import subprocess
+import sys
+import time
+import unittest
+from pathlib import Path
+
+input_file = "/ws/lambkin/input/hallway_localization"
+output_path = "/ws/lambkin/output/"
+num_iterations = 0
```
xaru8145: Shouldn't a good default value be at least 1?

src/lambkin/src/lambkin.py
```diff
@@ -0,0 +1,240 @@
...
+            sensor model and particle count.
+        iteration (int): The current iteration number.
+    """
+    variation_name = f"{variation['sensor_model']}_p{variation['num_particles']}"
+    base_dir = Path(output_path) / "benchmarking" / variation_name / f"iter_{iteration}"
+
+    p_beluga = beluga(
+        variation["sensor_model"],
+        variation["num_particles"],
+        "/ws/lambkin/input/map.yaml",
```
xaru8145: This is hardcoded. Use a global variable.

src/lambkin/src/lambkin.py
```diff
@@ -0,0 +1,240 @@
...
+
+    wait_for_processes([p_play], [p_beluga, p_record])
+
+
+def main():
+    """Main loop that orchestrates the benchmarking process."""
+    for variation in make_variations():
+        for it in range(num_iterations):
+            run_iteration(variation, it)
+            time.sleep(2)
```
xaru8145: Why is this sleep needed here?

src/lambkin/src/lambkin.py
```diff
@@ -0,0 +1,240 @@
...
+            variations.append(
+                {
+                    "sensor_model": sensor_model,
+                    "num_particles": particles,
+                }
+            )
+    return variations
+
+
+def run_iteration(variation: dict, iteration: int):
```
xaru8145:
```suggestion
def run_iteration(variation: dict, iteration: int) -> None:
```

src/lambkin/src/lambkin.py
```diff
@@ -0,0 +1,240 @@
...
+        p.terminate()
+    for p in termination_list:
+        try:
+            p.wait(timeout=5.0)
+        except subprocess.TimeoutExpired:
+            print("Forcing process termination...")
+            p.kill()
+
+
+def plot_ape_metrics(ape_path: str):
```
xaru8145:
```suggestion
def plot_ape_metrics(ape_path: str) -> None:
```

src/lambkin/src/lambkin.py
```diff
@@ -0,0 +1,240 @@
...
+        "--save_results",
+        str(ape_dir / "ape.zip"),
+    ]
+
+    subprocess.run(cmd_list, check=True)
+
+
+def wait_for_processes(
+    waitlist: list[subprocess.Popen], termination_list: list[subprocess.Popen]
+):
```
xaru8145:
```suggestion
) -> None:
```

src/lambkin/src/lambkin.py
```diff
@@ -0,0 +1,240 @@
...
+"""Lambkin library for executing and benchmarking Beluga AMCL."""
+
+import subprocess
+import sys
+import time
+import unittest
+from pathlib import Path
+
+input_file = "/ws/lambkin/input/hallway_localization"
+output_path = "/ws/lambkin/output/"
```
xaru8145: Where is the `reference_file`?

xaru8145 (changes requested):
First pass. Some general comments:

- Having `package.xml` and `CMakeLists.txt` inside `src/lambkin` is confusing. It makes the Python package look like a ROS package, which complicates installation via standard Python tools like `pip` or `uv`. I see 2 options here.
  - Have a ros2 package with launch and yaml examples (with its CMake and package.xml files).
  - Simply have a folder named `examples` where you leave any launch and yaml that the user will need to copy to their actual workspace/dev environment or slam package.
- The `src/lambkin/src` nesting is redundant and non-standard for Python.
- Putting `input` data (large `.db3` bags and `.pgm` maps) directly in the root is fine for a local MVP, but for an open-source project, these should be handled as external assets or placed in a specific examples/ or data/ directory.
- We are missing `src/lambkin/__init__.py`. Even if it starts empty, it is required to mark the directory as a regular Python package.

Regarding the PR description:

> Install the specific version of evo tools:
> ```
> pip install evo==1.34.3
> ```

Isn't this managed by `uv` in `lambkin` package dependencies? See `pyroject.toml` file.
Also, please explain in the PR description how the script manages folder creation based on iteration and variations.

teresa-ortega:
I've addressed some of the issues you mentioned. I'm still working on the remaining ones.

src/lambkin/src/lambkin.py
teresa-ortega: I've changed the variables names to avoid the confusion. Even though the "reference file" was the input file, now I explicitly define reference_bag_path and reference_map_math. 
Solved in b835167

src/lambkin/src/lambkin.py
teresa-ortega: Solved in b835167

src/lambkin/src/lambkin.py
teresa-ortega: Solved in b835167

src/lambkin/src/lambkin.py
teresa-ortega: Solved in b835167

src/lambkin/src/lambkin.py
teresa-ortega: that sleep was leftover from debugging and is not needed. I've removed it in b835167

lambkin_ros2/launch/beluga.launch.py
teresa-ortega: You're right, I changed the file name in b835167

src/lambkin/src/lambkin.py
teresa-ortega: You're right, I created the function in b835167

src/lambkin/src/lambkin.py
teresa-ortega: Solved in b835167

src/lambkin/test/test_process_management.py
teresa-ortega: Okay, I solved in b835167

src/lambkin/src/lambkin.py
xaru8145: I don't see the changes you mention in that commit.

teresa-ortega: Sorry, I thought that I had change that lines, now I will do a commit with that change.

src/lambkin/test/test_process_management.py
xaru8145: I see neither the pyproject.toml file modified to register the test folder nor the __init__.py file that addresses the comment. Please review.

src/lambkin/src/lambkin.py
teresa-ortega: Now I solved that in cc03561

src/lambkin/test/test_process_management.py
teresa-ortega: I solved that in the new commit b4fa905

src/lambkin/src/lambkin.py
teresa-ortega: Solved in cc03561

teresa-ortega: I changed the variable name to RECORD_TOPICS_INTERESTED. You can check in cc03561

src/lambkin/src/lambkin.py
teresa-ortega: You're right, I solved in cc03561

src/lambkin/src/lambkin.py
teresa-ortega: I changed in b835167

src/lambkin/lambkin.py
teresa-ortega: I implemented dry_mode in fc546bb

lambkin_ros2/launch/beluga.launch.py
teresa-ortega: I updated the launch file: I deleted the laser_topic parameter and created a new argument for the params file. If the user doesn't define a specific one, it takes the default file, which is the same one that Beluga uses.
You can see in b196d6c

xaru8145: There is a launch argument for the configuration file but the file is not passed to the beluga_node. If you are not sure how to do it please ping me.

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,306 @@
...
+import time
+from pathlib import Path
+
+reference_bag_path = (
+    "/ws/rosbags/magazzino_ros2_localization_only/bagfiles/hallway_localization/"
+)
+reference_map_path = "/ws/rosbags/maps/map.yaml"
+num_iterations = 1
+clock_rate = 10
+qos_file_path = "/ws/beluga/beluga_example/bags/qos_override.yaml"
```
xaru8145: Capitalize to signal they are global variables, just like the ones below.

lambkin_ros2/launch/beluga.launch.py
xaru8145, line 1: Rename `beluga_launch.py` to `beluga.launch.py`. It is typically the standard in ROS 2.

lambkin_ros2/params/default.ros2.yaml
```diff
@@ -0,0 +1,103 @@
+amcl:
```
xaru8145: Careful! This needs to match the node name as defined in the launch file, which is `beluga_amcl`. Otherwise the parameters won't be properly loaded. This can take a lot of time to debug and is typically overlooked.

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,306 @@
...
+                                   Defaults to False.
+
+    Returns:
+        subprocess.Popen: The running launch process.
+    """
+    cmd_list = [
+        "ros2",
+        "launch",
+        "lambkin",
+        "lambkin_launch.py",
```
xaru8145: The package name here is a bit tricky as it will depend on where the user places the example. We can have beluga as a placeholder. 
```suggestion
        "beluga",
        "beluga.launch.py",
```
At this point of development, to simplify, I think it might be worth having the examples folder as a ros2 package with a CMake and package. You can rename examples to signal is related to ros2. And add a simple dockerfile like we discussed. So the beluga placeholder above would then be the package name.

If in next phases we decide to remove the ros 2 package or add it in another repo is fine. It is not set in stone.

examples/beluga_launch.py
```diff
@@ -0,0 +1,77 @@
...
+    Declares launch arguments for the map path, laser model type,
+    and maximum particles,and configures the required ROS 2 nodes:
+    beluga_amcl, map_server, and lifecycle_manager.
+
+    Returns:
+        LaunchDescription: The complete ROS 2 launch description object.
+    """
+    config_file_arg = DeclareLaunchArgument(
+        "beluga_config_path",
+        default_value="/example/defaul_params.ros.yaml",  # Opcional por defecto
```
xaru8145: Comments should be in English (if any).

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,306 @@
...
+    for p in termination_list:
+        if p is not None:
+            p.terminate()
+    for p in termination_list:
+        if p is not None:
+            try:
+                p.wait(timeout=5.0)
+            except subprocess.TimeoutExpired:
+                print("Forcing process termination...")
+                p.kill()
```
xaru8145, line 322: Please add code comments.

Why do you run 2 for loops through `termination_list`? Not saying it is wrong, but argue why is done like this.

Consider moving the hardcoded timeout value of 5.0 to a global variable e.g. `PROCESS_TERMINATION_TIMEOUT`. In this way it is easier for the user to see all variables that can be changed at a single glance rather than having to go through the code to find hidden hardcoded values.

xaru8145 (changes requested):
Good job! Getting there but there are still some things to polish.

Review PR description. There are still references to lambkin launch and the execution of the script has a typo (`//`):
>  ```
> python3 src/lambkin//lambkin.py
> ```

teresa-ortega:
Fix for test import issues (Commit c60ecbc)

* Fixed the failing tests caused by import errors from the lamkin module.

* Added an __init__.py file to the /lamkin/ directory to make it a recognizable Python package, enabling correct imports in the test files.

* Ensured no __init__.py is present in the tests directory. Since it is not a package, having an init file is unnecessary and can cause unexpected behavior when running pytest.

src/lambkin/lambkin.py
teresa-ortega: I used two separate loops to ensure the processes shut down in parallel. In the first loop, I send the SIGTERM signal to all the processes, but since I don't know right away if they handled the signal correctly, the second loop is there to make sure they actually terminate. If a process has already terminated, p.wait() passes immediately. If not, I give the process 5 seconds to finish; if that time runs out, I force-kill the process.

xaru8145: Great, thanks for the detailed explanation. That makes sense. Please add a **brief** code comment explaining this design decision.

examples/beluga_launch.py
teresa-ortega: Solved in fd2879b

lambkin_ros2/launch/beluga.launch.py
teresa-ortega: Solved in fd2879b

src/lambkin/lambkin.py
teresa-ortega: Solved in fd2879b

src/lambkin/lambkin.py
teresa-ortega: You can see the comments and the global variable in 2c62de0

src/lambkin/lambkin.py
teresa-ortega: I've already created the ROS 2 package with the CMakeLists and package.xml files, and fixed the naming confusion in commit 86a569e if you want to check it

lambkin_ros2/params/default.ros2.yaml
teresa-ortega: Solved in 86a569e

lambkin_ros2/launch/beluga.launch.py
teresa-ortega: Solved in 86a569e

lambkin_ros2/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+cmake_minimum_required(VERSION 3.16)
+project(lambkin_ros2)
+find_package(ament_cmake REQUIRED)
+find_package(ament_cmake_python REQUIRED)
+
+
+
+install(
+  DIRECTORY
+    launch
```
xaru8145, line 12: Install also `params` folder please.

lambkin_ros2/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+
+cmake_minimum_required(VERSION 3.16)
+project(lambkin_ros2)
+find_package(ament_cmake REQUIRED)
+find_package(ament_cmake_python REQUIRED)
```
xaru8145: Review since ament_cmake_python is not needed and use defensive programming by adding as REQUIRED the exec depends (altough not strictly needed).

lambkin_ros2/package.xml
```diff
@@ -0,0 +1,23 @@
...
+
+
+  <exec_depend>python3-numpy</exec_depend>
+  <exec_depend>python3-pandas</exec_depend>
+  <exec_depend>python3-tk</exec_depend>
+
+
+  <exec_depend>ros2bag</exec_depend>
+  <exec_depend>rosbag2_transport</exec_depend>
+  <exec_depend>rosbag2_storage_default_plugins</exec_depend>
```
xaru8145: Please review dependencies. As discussed via call, add launch related ones and also the packages that are run as part of the launch.

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,310 @@
+#!/usr/bin/env python3
+"""Lambkin library for executing and benchmarking Beluga AMCL."""
```
xaru8145: Not only Beluga, any SLAM algorithm. Beluga is just an example and use case.

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,310 @@
...
+import subprocess
+import time
+from pathlib import Path
+
+REFERENCE_BAG_PATH = (
+    "/ws/rosbags/magazzino_ros2_localization_only/bagfiles/hallway_localization/"
+)
+REFERENCE_MAP_PATH = "/ws/rosbags/maps/map.yaml"
+NUM_ITERATIONS = 1
+CLOCK_RATE = 10
```
xaru8145:
```suggestion
RATE = 1
```

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,310 @@
...
+QOS_FILE_PATH = "/ws/beluga/beluga_example/bags/qos_override.yaml"
+
+
+LASER_MODELS = ["likelihood_field", "beam"]
+NUM_PARTICLES = [1, 10, 100, 1000, 10000]
+CLOCK_RATE_OPTION = ["--clock-rate", str(CLOCK_RATE)]
```
xaru8145:
```suggestion
RATE_OPTION = ["--rate", str(RATE)]
```

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,310 @@
...
+QOS_OPTION = ["--qos-profile-overrides-path", QOS_FILE_PATH]
+CLOCK_OPTION = "--clock"
```
xaru8145: Review as this option is unused and needed for the ros2 bag play.

xaru8145 (changes requested):

src/lambkin/lambkin.py
teresa-ortega: Solved in f8fadf0

src/lambkin/lambkin.py
teresa-ortega: Solved in f8fadf0

src/lambkin/lambkin.py
teresa-ortega: Made the --clock option mandatory in the ros_bag_play method in f8fadf0

src/lambkin/lambkin.py
teresa-ortega: Solved in f8fadf0

lambkin_ros2/package.xml
teresa-ortega: Solved in 87056cf

lambkin_ros2/CMakeLists.txt
teresa-ortega: Solved in 87056cf

lambkin_ros2/CMakeLists.txt
teresa-ortega: Solved in 87056cf

src/lambkin/lambkin.py
teresa-ortega: I've created the Dockerfile and docker-compose as we discussed during our call. I've tested them, and they seem to be working correctly in 67c8eed

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,313 @@
+#!/usr/bin/env python3
+"""Lambkin library for executing and benchmarking."""
```
xaru8145: for executing and benchmarking what?
```suggestion
"""Lambkin library for executing and benchmarking Localization and Mapping algorithms."""
```

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,313 @@
...
+    Args:
+        input_path (str): The path to the input bag file.
+        options (list[str]): Additional options for playback (e.g., clock rate, QoS).
+        dry_mode (bool, optional): If True, prints the command without executing it.
+                                   Defaults to False.
+
+    Returns:
+        subprocess.Popen: The running ros2 bag play process.
+    """
+    cmd_list = ["ros2", "bag", "play", input_path, CLOCK_OPTION] + options
```
xaru8145: Although we will mostly replay with clock option you have to think about this function as something generic the user can use as pleased. So if the user does not want to do a rosbag play with clock, we have to allow that. Therefore, do not hardcode it here and add it in `run_iteration`, where we call this function.
```suggestion
    cmd_list = ["ros2", "bag", "play", input_path] + options
```

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+
+RUN apt-get update && apt-get install -y \
+    curl \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /ws
+
+
+COPY lambkin_ros2/ /ws/lambkin_ros_ws/src/lambkin_ros2/
```
xaru8145: Careful here! You are copying the source code and in the docker compose you are mounting it. So you have 2 ros2 workspaces:
- lambkin_ros2 (mounted by docker compose).
- lambkin_ros_ws (created in the dockerfile): this one remains static as you develop as it is copied at the moment you build the image.
```bash
root@nebula:/ws# ls
lambkin_ros2  lambkin_ros_ws  pyproject.toml  reference  src  uv.lock
```

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+COPY /src/lambkin/ /ws/src/lambkin/
+COPY pyproject.toml uv.lock ./
```
xaru8145: Why copy this if you are not installing any project dependencies in the image and you are mounting it in the docker compose?

docker/Dockerfile
xaru8145, line 1: Please move Dockerfile and docker compose to a `docker` folder.

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,313 @@
...
+        subprocess.Popen: The running ros2 bag record process.
+    """
+    bag_dir = Path(output_path) / "bag"
+    bag_dir.mkdir(parents=True, exist_ok=True)
+
+    if bag_dir.exists():
+        import shutil
+
+        shutil.rmtree(bag_dir)
+        print(f"Borrando carpeta antigua en {bag_dir}...")
```
xaru8145: Use English please.

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,313 @@
...
+"""Lambkin library for executing and benchmarking."""
+
+import subprocess
+import time
+from pathlib import Path
+
+REFERENCE_BAG_PATH = "/ws/reference/hallway_localization"
+REFERENCE_MAP_PATH = "/ws/reference/map.yaml"
+NUM_ITERATIONS = 1
+RATE = 100
```
xaru8145: Shouldn't rate be 1?

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,313 @@
...
+
+
+def main():
+    """Main loop that orchestrates the benchmarking process."""
+    for variation in make_variations():
+        for it in range(NUM_ITERATIONS):
+            run_iteration(
+                variation, it, REFERENCE_MAP_PATH, REFERENCE_BAG_PATH, DRY_MODE
+            )
+    # evo_ape(REFERENCE_TUM_PATH, "../benchmarking_results/")
```
xaru8145: What is the purposes of having this code call commented? If it's not used, remove.  Otherwise, uncomment.

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,313 @@
...
+    )
+    time.sleep(3)
+
+    p_play = ros_bag_play(
+        reference_bag_path, RATE_OPTION + QOS_OPTION, dry_mode=dry_mode
+    )
+
+    p_record = ros_bag_record(
+        str(base_dir), RECORD_TOPICS_INTERESTED, dry_mode=dry_mode
+    )
```
xaru8145, line 401: If we run the record after play there is a risk we loose messages. Invert order

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,313 @@
+#!/usr/bin/env python3
+"""Lambkin library for executing and benchmarking."""
+
+import subprocess
+import time
+from pathlib import Path
+
+REFERENCE_BAG_PATH = "/ws/reference/hallway_localization"
+REFERENCE_MAP_PATH = "/ws/reference/map.yaml"
+NUM_ITERATIONS = 1
```
xaru8145: Add a global variable `LOG_PATH` where we can store the logs generated by redirecting background subprocesses to a file so that we only hardcode the file name. For its default value, it is a good idea to reuse the `log/` folder ros2 automatically creates when building (`colcon build`), which I already suggested to add to gitignore so we will not push this files to git.
```suggestion
REFERENCE_BAG_PATH = "/ws/reference/hallway_localization"
REFERENCE_MAP_PATH = "/ws/reference/map.yaml"
LOG_PATH = "/ws/log"
NUM_ITERATIONS = 1
```

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,313 @@
...
+    cmd_list = [
+        "ros2",
+        "launch",
+        "lambkin_ros2",
+        "beluga.launch.py",
+        f"map_path:={map_path}",
+        f"laser_model_type:={sensor_model}",
+        f"max_particles:={num_particles}",
+    ]
+    return execute_background_process(cmd_list, dry_mode, log_file="beluga.log")
```
xaru8145, line 206: Related to comment above:
```suggestion
    beluga_log_path = LOG_PATH / "beluga.log"
    return execute_background_process(cmd_list, dry_mode, log_file=str(beluga_log_path))
```

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,313 @@
...
+    variation_name = f"{variation['sensor_model']}_p{variation['num_particles']}"
+    base_dir = Path(RESULTS_PATH) / variation_name / f"iter_{iteration}"
+
+    p_beluga = beluga(
+        variation["sensor_model"],
+        variation["num_particles"],
+        map_reference_path,
+        dry_mode=dry_mode,
+    )
+    time.sleep(3)
```
xaru8145: Suggestion: parameterize sleep time.

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+    network_mode: "host"
+
+    stdin_open: true
+    tty: true
+
+    volumes:
+      - ./lambkin_ros2:/ws/lambkin_ros2
+      - ./src/lambkin:/ws/src/lambkin
+      - ./pyproject.toml:/ws/pyproject.toml
+      - ./uv.lock:/ws/uv.lock
```
xaru8145: As mentioned in the review comment in detail, we mount the entire repo to the workspace folder.
```suggestion
      - .:/ws
```

Dockerfile
xaru8145: We are mounting the entire repo which includes python source code, uv files, pyproject files and ros2 package. So remove all the copy statements.

Dockerfile
xaru8145: I left a comprehensive comment on the review title.

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,313 @@
+#!/usr/bin/env python3
+"""Lambkin library for executing and benchmarking."""
+
+import subprocess
+import time
+from pathlib import Path
+
+REFERENCE_BAG_PATH = "/ws/reference/hallway_localization"
```
xaru8145: Better mount outside of the workspace as now we mount it entirely and it will be tracked by git. So rather than adding a new folder to exclude in gitignore I prefer to just simply mount it outside of the workspace, which also helps organize better (workspace vs rosbags -> clear separation)
```suggestion
REFERENCE_BAG_PATH = "/rosbags/reference/hallway_localization"
```

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,313 @@
...
+REFERENCE_MAP_PATH = "/ws/reference/map.yaml"
```
xaru8145: Same comment as above:

```suggestion
REFERENCE_MAP_PATH = "/rosbags/reference/map.yaml"
```

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,313 @@
...
+QOS_FILE_PATH = "/ws/reference/qos_override.yaml"
+
+
+LASER_MODELS = ["likelihood_field", "beam"]
+NUM_PARTICLES = [1, 10, 100, 1000, 10000]
+RATE_OPTION = ["--rate", str(RATE)]
+QOS_OPTION = ["--qos-profile-overrides-path", QOS_FILE_PATH]
+CLOCK_OPTION = "--clock"
+RECORD_TOPICS_INTERESTED = ["/pose", "/tf", "/tf_static"]
+RESULTS_PATH = "/ws/src/lambkin/benchmarking_results"
```
xaru8145: Better mount outside of the workspace as now we mount it entirely and it will be tracked by git. So rather than adding a new folder to exclude in gitignore I prefer to just simply mount it outside of the workspace, which also helps organize better (workspace vs results -> clear separation)
```suggestion
RESULTS_PATH = "/benchmarking_results"
```

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+      - ${HOME}/ekumen/rosbags/reference_files:/ws/reference
```
xaru8145: Related to comments above:
```suggestion
      - ${HOME}/ekumen/rosbags/reference_files:/rosbags/reference
```

xaru8145 (changes requested):
PTAL.

I changed my mind on how to mount the ros2 package vs the lambkin python library after playing with it a bit. The suggestion I made earlier to split both was very prone to error.
I prefer to mount all the repo to the workspace folder and directly run `uv` and `colcon` commands there. This requires to update the `.gitignore` file so to ignore ros2 related folders (`install/`, `build/` and `log/`).

Use `--topics` flag when playing the rosbag as the API suggests:
```bash
[WARN] [ros2bag]: Positional "topics" argument deprecated. Please use optional "--topics" argument instead.
```

Also, update the PR description with all the needed steps and its commands, as it has obsolete commands like the one below:
> ```
> colcon build --symlink-install --packages-ignore beluga_vdb
> ```
Steps may include:
- Build docker image.
- Run the docker image.
- Install dependencies using rosdep.
- Build the workspace.
- Source the install space.
- ...

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,313 @@
...
+                                   Defaults to False.
+
+    Returns:
+        subprocess.Popen: The running ros2 bag record process.
+    """
+    bag_dir = Path(output_path) / "bag"
+    bag_dir.mkdir(parents=True, exist_ok=True)
+
+    if bag_dir.exists():
+        import shutil
```
xaru8145: Why define the import here instead of the beginning of the file?

Dockerfile
teresa-ortega: I agree with mounting the entire repo to avoid errors. However, to make colcon build work properly, should we move lambkin_ros2 into the src folder? Let me know where you'd prefer to place it with this new setup.

xaru8145: No need to. colcon is capable of only building ros2 packages. And does not need to look into a src folder. It is the standard way of builidng workspaces but for this phase let's keep it simple to not mix python source code and ros source code

src/lambkin/lambkin.py
teresa-ortega: Solved in d61b39b

src/lambkin/lambkin.py
teresa-ortega: Solved in d61b39b

src/lambkin/lambkin.py
teresa-ortega: Yes, of course. Solved in d61b39b

src/lambkin/lambkin.py
teresa-ortega: Solved in d61b39b

src/lambkin/lambkin.py
teresa-ortega: Solved in d61b39b

src/lambkin/lambkin.py
teresa-ortega: Solved in d61b39b

src/lambkin/lambkin.py
teresa-ortega: Solved in d61b39b

src/lambkin/lambkin.py
teresa-ortega: Perfect, solved in d61b39b

src/lambkin/lambkin.py
teresa-ortega: Solved in d61b39b

src/lambkin/lambkin.py
teresa-ortega: Solved in d61b39b

src/lambkin/lambkin.py
teresa-ortega: Solved in d61b39b

src/lambkin/lambkin.py
teresa-ortega: Solved in d61b39b

docker-compose.yml
teresa-ortega: Solved in e5672f0

Dockerfile
teresa-ortega: Solved in e5672f0

src/lambkin/lambkin.py
teresa-ortega: Solved in e5672f0

docker-compose.yml
teresa-ortega: Solved in e5672f0

docker/Dockerfile
teresa-ortega: Solved in e5672f0

Dockerfile
teresa-ortega: You are correct. As discussed on our call, we don't need to install any project dependencies. I have resolved this in commit e5672f0

Dockerfile
teresa-ortega: Solved in e5672f0

src/lambkin/lambkin.py
xaru8145: Why is rate back to 100?

xaru8145 (changes requested):
Almost there.

Despite not having any folder in benchmarking results I get the msg: 
`Removing existing directory at /benchmarking_results/likelihood_field_p1/iter_0/bag...`
which is confusing.

You can see if I list the `benchmarking_results` directory, there is nothing. And in contrast, I still get the message saying we remove it.
```bash
root@nebula:/ws# ls /benchmarking_results/
root@nebula:/ws# uv run src/lambkin/lambkin.py
Running iteration 0 with likelihood_field_p1
Removing existing directory at /benchmarking_results/likelihood_field_p1/iter_0/bag...
```

I am also getting several ros2 node with the same name. What do you think is the reason?

```bash
root@nebula:/benchmarking_results# ros2 node list
WARNING: Be aware that there are nodes in the graph that share an exact name, which can have unintended side effects.
/beluga_amcl
/lifecycle_manager_localization
/lifecycle_manager_localization
/lifecycle_manager_localization
/lifecycle_manager_localization
/lifecycle_manager_localization
/lifecycle_manager_localization
/lifecycle_manager_localization
/lifecycle_manager_localization
/map_server
/map_server
/map_server
/rosbag2_player
/rosbag2_recorder
```

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,440 @@
...
+            nodes.append(node)
+
+    if not nodes:
+        print("No orphaned ROS2 nodes found.")
+        return
+
+    print(f"Found orphaned nodes: {nodes}")
+
+    for node in nodes:
+        subprocess.run(["pkill", "-f", node.lstrip("/")], capture_output=True)
```
xaru8145, line 52: Might consider using -9 for a more aggressive termination.

src/lambkin/lambkin.py
```diff
@@ -0,0 +1,440 @@
...
+APE_TOPICS_INTERESTED = "/pose"
+
+
+def kill_ros2_nodes() -> None:
+    """Kill any orphaned ROS2 nodes before starting the benchmark."""
+    subprocess.run(["ros2", "daemon", "stop"], capture_output=True)
+    subprocess.run(["ros2", "daemon", "start"], capture_output=True)
+    time.sleep(1)
+
+    result = subprocess.run(["ros2", "node", "list"], capture_output=True, text=True)
```
xaru8145, line 38: Could this be run with `execute_foreground_process`? Or is there an argument required not forwarded to the function?

xaru8145 (approved):
LGTM. Good job 💪🏼 

Still seeing evo warning message:
```bash
/ws/.venv/lib/python3.12/site-packages/seaborn/axisgrid.py:123: UserWarning: The figure layout has changed to tight
  self._figure.tight_layout(*args, **kwargs)
```
and the axis layout from some plots overlap, potentially due to this warning. Fine to address this and the comments in a follow up PR. Let's merge this asap.

Image

src/lambkin/lambkin.py
teresa-ortega: Yes, it's possible. We just need to update the execute_foreground_process method to include a new parameter that captures the output.

src/lambkin/lambkin.py
teresa-ortega: I was running some tests with kill -9 and pkill -f, and both have some issues. In my opinion, the nodes seem to disappear faster with pkill -f . Also, the downside of kill -9 is that we have to run an extra command first just to catch the process ID (PID).

PR #134, opened 2026-03-12
Phase 3: core/ctx/context and core/ctx/source

teresa-ortega:
### Proposed changes

This PR introduces context.py and the benchmark Source class, which encapsulate all path and parameter information for a single benchmark run. A Context instance is created once per (variation, iteration) pair and passed into the algorithm, replacing the previous pattern of passing loose parameters across functions.

#### Type of change

### Checklist

### Additional comments


src/lambkin/core/ctx/context.py
```diff
@@ -32,13 +32,27 @@ class VariationInfo:
...
+        **options : Any
+            Option names (already converted from --flag-name to flag_name)
+            and their values as parsed by click.
+        """
+        for key, value in options.items():
+            setattr(self, key, value)
+
+    def as_dict(self) -> dict[str, Any]:
+        """Return all options as a plain dict."""
+        return dict(self.__dict__)
```
xaru8145: The current OptionsInfo class manually replicates what types.SimpleNamespace does out of the box. Example [here](https://dev.to/taqkarim/extending-simplenamespace-for-nested-dictionaries-58e8).

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,254 @@
...
+        # ctx.variation
+        self.variation = VariationInfo(
+            sensor_model=variation["sensor_model"],
+            num_particles=variation["num_particles"],
+        )
+
+        # ctx.options
+        self.options = OptionsInfo(**options)
+
+        # ctx.source — may be overridden by @nomida.input
```
```suggestion
        # ctx.source — may be overridden by @nominal.input
```

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,254 @@
...
+        self.source = SourceInfo(path=Path(source_path) if source_path else Path())
+
+        # ctx.inputs — may be overridden by @nomida.input
```
```suggestion
        # ctx.inputs — may be overridden by @nominal.input
```

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,254 @@
...
+from pathlib import Path
+from typing import Any
+
+
+@dataclass
+class VariationInfo:
+    """Algorithm parameters for this benchmark variation."""
+
+    sensor_model: str
+    num_particles: int
```
xaru8145: What happens if we want to use another variation e.g. `motion_model`? Please keep the SDK parameters and variations agnostic. Consider using `SimpleNamespace` (like suggested with options).

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,254 @@
...
+        variation : dict
+            Algorithm parameters for this run. Must contain:
+            - ``sensor_model`` (str): e.g. ``"beam"`` or ``"likelihood_field"``.
+            - ``num_particles`` (int): number of particles for the algorithm.
+        iteration : int
+            Zero-based index of the current repetition within this variation.
+            Determines the ``iter_<N>`` subfolder under the variation directory.
+        source_path : Path or str
+            Filesystem path to the ROS source package under test
+            (e.g. ``/opt/ros/overlay/amcl``).
```
xaru8145: Why do we need SourceInfo? LAMBKIN needs to be ROS agnostic.

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,254 @@
...
+        dataset_path : Path or str
+            Path to the rosbag file used as input by the algorithm
+            (e.g. ``data/rosbags/run_01.bag``).
+        output_dir : Path or str
+            Root directory where all benchmark results are written.
+            The variation and iteration subfolders are created inside it.
+        option : dict
+            ROS runtime options. Recognised keys:
```
xaru8145: Not necessarily ROS.

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,254 @@
...
+
+        # ctx.iteration
+        self.iteration = iteration
+
+        # ctx.output
+        base = Path(output_dir)
+        variation_dir = base / _variation_folder_name(
+            self.variation.sensor_model,
+            self.variation.num_particles,
+        )
```
xaru8145: This is hardcoded. We do not know what variations the user will prompt. Related to the comment in `VariationInfo` class.

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,254 @@
...
+
+@dataclass(frozen=True)
+class InputsInfo:
+    """Input data paths."""
+
+    dataset: Path
+
+    def __post_init__(self) -> None:
+        """Convert dataset to a Path object after dataclass construction."""
+        object.__setattr__(self, "dataset", Path(self.dataset))
```
xaru8145: Why use `frozen` and `__setattr__` to intialize the input path?

test/core/test_context.py
xaru8145, line 1: Tests will likely need to be revisited once the context is updated to address the review comments.

xaru8145 (changes requested):
First pass. PTAL but great job with this! Keep it up 💪🏼

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in 6935aef

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in b2aca77

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in b2aca77

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in b2aca77

src/lambkin/core/ctx/context.py
teresa-ortega: I've added source path since it appears in the API

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in 4e7d562

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,230 @@
...
+from types import SimpleNamespace
+from typing import Any
+
+
+@dataclass
+class VariationInfo:
+    """Algorithm parameters for this benchmark variation."""
+
+    sensor_model: str
+    num_particles: int
```
xaru8145: Great refactor to SimpleNamespace. Note that the `VariationInfo` dataclass is now unused and can be removed.

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,230 @@
...
+
+    def add_variation(self, variation: dict) -> None:
+        """Set variation parameters from a dict onto ctx.variation."""
+        for key, value in variation.items():
+            setattr(self.variation, key, value)
+
+    def add_options(self, options: dict) -> None:
+        """Set option parameters from a dict onto ctx.options."""
+        for key, value in options.items():
+            setattr(self.options, key, value)
```
xaru8145: Will `add_options` and `add_variation` actually be used? The options and variations are being instantiated in the constructor and I believe they will actually bet set only one in the options and benchmark decorator.

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,230 @@
...
+            (e.g. ``data/rosbags/run_01.bag``). Defaults to an empty path.
+
+        output_dir : Path or str
+            Root directory for all benchmark results.
+            Variation and iteration subfolders are created inside it.
+
+        options : dict
+            Runtime options, as defined by the user.
+            All key-value pairs are exposed as attributes on ``ctx.options``.
+        """
```
xaru8145, line 130: Improve the documentation. See example here https://realpython.com/documenting-python-code/#class-docstrings

Your class `OutputInfo` is a good example.

Use Attributes for class documentation, and Parameters for init.

Move the high-level 'what' (the public Attributes and Methods, if any) to the class docstring to define the object's interface. Keep the __init__ docstring focused strictly on the 'How', by documenting the raw Parameterse. Documentation of Methods is optional, but if done, only for public methods.

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,230 @@
...
+        return self._make(self.iteration_dir / "metrics")
+
+
+def _variation_folder_name(variation: SimpleNamespace) -> str:
+    parts = []
+    for value in vars(variation).values():
+        if isinstance(value, float) and value.is_integer():
+            value = int(value)
+        parts.append(str(value))
+    return "_".join(parts)
```
xaru8145: I am not convinced about this. From your test example:
This variation -> "sensor_model": "beam", "num_particles": 10 generates folder  `beam_10`. But what if we have more variations that use number like `max_beams` 100. Then how do I know if 10 is for `num_particles` or  `max_beams`?

Aside from this, I believe we have to re-think this a bit more. What happens when we are changing many set of parameters?
- sensor_model: beam
- motion_model: omni
- num_particles: 100
- map_resolution: 0.05
- max_beams: 100

The folder name would be `beam_omni_100_005_100`. Imagine if we have 10 parameters or we also add the parameter name before the value `sensor_model_beam_motion_model_omni_np...`. You can see the folder name can get very long. And Linux imposes limits on folder (and file) name lengths.

An idea we can consider or explore is using a hash value and leaving a metadata file that associates the hash with the variation with parameter and value. There could be other options. In any case, we can discuss this in the next meeting.

xaru8145 (changes requested):
PTAL

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in cf23651

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in cf23651

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in cf23651

src/lambkin/core/ctx/context.py
teresa-ortega: Using frozen is to make sure the input can't be re declared. To make it simpler, since it's not really necessary anyway, I've solved it a different way in commit cf23651

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in cf23651

test/core/test_context.py
teresa-ortega: Solved in aadd392

src/lambkin/core/ctx/context.py
teresa-ortega: Created a new folders structure in aadd392

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,185 @@
...
+            f"Context(\n"
+            f"  variation    = {vars(self.variation)},\n"
+            f"  iteration    = {self.iteration},\n"
+            f"  source       = {source.path if source else 'not set'},\n"
+            f"  inputs       = {inputs.dataset if inputs else 'not set'},\n"
+            f"  options      = {vars(self.options)},\n"
+            f"  variation_dir= {self.output.variation_dir},\n"
+            f"  iteration_dir= {self.output.iteration_dir}\n"
+            f")"
+        )
```
xaru8145, line 179: This is currently broken as it references dataset. Also `self.output.variation_dir` and `self.output.iteration_dir` no longer exist.

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,185 @@
...
+    """Output paths for this variation + iteration.
+
+    Folders are created lazily — only when the path is first accessed.
+    This means a bag/ folder is only created if the user accesses
+    ctx.output.bag_dir, and same for ape/ and any other subfolder.
+
+    Attributes:
+    ----------
+    variation_dir:
+        Root folder for this variation (e.g. results/beam_p100/).
```
xaru8145: I think this reflects old folder naming.

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,185 @@
...
+        Created eagerly on Context instantiation.
+    iteration_dir:
+        Folder for the current iteration (e.g. results/beam_p100/iter_0/).
```
xaru8145: I think this reflects old folder naming.

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,185 @@
...
+        Created eagerly on Context instantiation.
+    bag_dir:
+        Subfolder for rosbag output (iter_N/bag/).
+        Created on first access.
+    ape_dir:
+        Subfolder for APE results (iter_N/ape/).
+        Created on first access.
```
xaru8145: Do we really need to have a folder for bag and a folder for ape? Could it make sense to leave everything on `var_x/iter_y` folder?

On a similar note, what if we do not want to run APE, but want to run RPE? What if instead we do not want to run evo but cpu and ram metrics?

We have to re-think this too a bit.

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,185 @@
...
+
+        Only the base folders (variation_dir and iteration_dir) are stored
+        at construction time. Subfolders like bag/ and ape/ are created
+        lazily on first access via their respective properties.
+
+        Parameters
+        ----------
+        variation_dir : Path
+            Root output folder for this variation
+            (e.g. results/beam_p100/).
```
xaru8145: I think this reflects old folder naming.

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,185 @@
...
+        iteration_dir : Path
+            Output folder for the current iteration
+            (e.g. results/beam_p100/iter_0/).
```
xaru8145: I think this reflects old folder naming.

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,185 @@
...
+            Controls the "var_<N>" subfolder name under the output directory,
+        """
+        # ctx.variation
+        self.variation = SimpleNamespace(**variation)
+
+        # ctx.options
+        self.options = SimpleNamespace(**options)
+
+        # ctx.source — may be overridden by @nominal.input
+        self.source = Path.cwd()
```
xaru8145: `Path.cwd()` returns the folder where the user is (the terminal location), not where the script lives.
- If I am in `/home/user/` and I run `python benchmarks/test.py`, `cwd()` returns `/home/user/`.
- Your code will then look for `beluga.launch.xml` in `/home/user/`, and it won't find it because the file is actually in `/home/user/benchmarks/`.

Also, the comment above is not correct (`# ctx.source — may be overridden by @nominal.input`) as source should not be overridden.

According to the SKD,` ctx.source` is supposed to be an object describing the benchmark source, and `ctx.source.path` is supposed to be a `pathlib.Path` to the source file of the benchmark. In the current implementation, the source is just the path and is accessed directly with `ctx.source`. Before moving forward, we need to understand what else other than the Path can live in a Source object to see whether it should be constructed by a decorator and passed to the Context constructor or built in the same Context constructor.

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,185 @@
...
+    output) from the given parameters and automatically creates the
+    required output folders on disk.
+
+    Attributes:
+    ----------
+    variation:
+        Namespaced algorithm parameters for this run.
+        All key-value pairs from the variation dict are exposed as attributes.
+    source:
+        Path to the directory where the script is being executed.
```
xaru8145: It contains the path but I am not sure it is only the path (I will figure it out and get back to you asap).
```suggestion
        Describes the benchmark source.
```

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,185 @@
...
+        Automatically set at instantiation time.
+    inputs:
+        Namespaced input/dataset information (path to the rosbag file).
```
xaru8145: Generalize from ros
```suggestion
        Namespaced input/dataset information (e.g. path to the mcap file).
```

test/core/test_context.py
xaru8145, line 1: Add a test case (e.g. `test_inputs_can_be_dynamically_populated`) to verify that `ctx.inputs` correctly accepts dynamic attribute assignments via `setattr()`. This is necessary to verify the behavior required by the `@nominal.input` system, which injects data post-instantiation, unlike options which are populated at construction.

xaru8145 (changes requested):
Some changes are required. And, I have some questions:
- What happens if we provide the same key for 2 variations e.g. `sensor_model` twice? Would it throw an exception?
- What happens if we provide the same values for a given key? For example,  `num_particles: [10, 10]`?

src/lambkin/core/ctx/context.py
xaru8145: @teresa-ortega We expect the `Source` object to have more data, we just do not need to model it now. For now, `Source` should be an object with just a `Path` attribute so that it can be extended with more attributes and functionality in the future.

If we make Source `directly` a `Path` object we are limiting its extension in the future. Consider passing the entire Source object upon Context construction.

And let's think about whether its worth it to have a `source.py` file with the `Source` class.

src/lambkin/core/ctx/context.py
teresa-ortega: The bag and metrics subfolders aren't strictly necessary for the benchmark execution. They just exist to provide better folder organization for the user.

src/lambkin/core/ctx/context.py
teresa-ortega: I fixed the dataset issue, but the variation and iteration are defined in the OutputInfo class. 
15c5632

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in 15c5632

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in 15c5632

src/lambkin/core/ctx/context.py
teresa-ortega: The reference to ape was a typo. I meant to refer to metrics, not ape
15c5632

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in 15c5632

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in 15c5632

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in 15c5632

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in 15c5632

src/lambkin/core/ctx/context.py
xaru8145: True! Thanks for pointing that out

test/core/test_source.py
```diff
@@ -0,0 +1,38 @@
...
+
+from pathlib import Path
+
+from lambkin.core.ctx.source import Source
+
+
+def test_source_path_is_set_correctly():
+    """ctx.source.path reflects the path passed at construction."""
+    source = Source(path=Path(__file__))
+    assert source.path == Path(__file__)
```
xaru8145: You use `Path(__file__)` twice. To make the tests cleaner and more maintainable define it as a variable.
```suggestion
def test_source_path_is_set_correctly():
    """ctx.source.path reflects the path passed at construction."""
    test_file = Path(__file__)
    source = Source(path=test_file)
    assert source.path == test_file
```

test/core/test_source.py
```diff
@@ -0,0 +1,38 @@
...
+    """Source converts a string path to a Path object."""
+    source = Source(path="/home/user/benchmarks/test.py")
+    assert isinstance(source.path, Path)
+    assert source.path == Path("/home/user/benchmarks/test.py")
+
+
+def test_source_repr():
+    """Source.__repr__ returns a human-readable string."""
+    source = Source(path=Path("/home/user/benchmarks/test.py"))
+    assert repr(source) == "Source(path=/home/user/benchmarks/test.py)"
```
xaru8145:
```suggestion
def test_source_repr():
    """Source.__repr__ returns a human-readable string."""
    test_path = Path("/home/user/benchmarks/test.py")
    source = Source(path=test_path)
    assert repr(source) == f"Source(path={test_path})"
```

test/core/test_source.py
```diff
@@ -0,0 +1,38 @@
...
+    """ctx.source.path reflects the path passed at construction."""
+    source = Source(path=Path(__file__))
+    assert source.path == Path(__file__)
+
+
+def test_source_path_is_converted_to_path():
+    """Source converts a string path to a Path object."""
+    source = Source(path="/home/user/benchmarks/test.py")
+    assert isinstance(source.path, Path)
+    assert source.path == Path("/home/user/benchmarks/test.py")
```
xaru8145: Defining variables in one place that are re-used helps with maintainability.
```suggestion
def test_source_path_is_converted_to_path():
    """Source converts a string path to a Path object."""
    test_path_str = "/home/user/benchmarks/test.py"
    source = Source(path=test_path_str)
    assert isinstance(source.path, Path)
    assert source.path == Path(test_path_str)
```

test/core/test_source.py
xaru8145, line 1: Suggestion: add a test to simulate and validate the usage of source path anchor in the nominal function. For example:

```py
def test_source_parent_anchor():
    """Verify that source.path.parent can be used to anchor relative paths."""
    test_path = Path("/home/user/benchmarks/my_benchmark.py")
    source = Source(path=test_path)
    dataset_path = source.path.parent / "datasets" / "output.mcap"
    
    assert dataset_path == test_path.parent / "datasets" / "output.mcap"
```

test/core/test_context.py
```diff
@@ -0,0 +1,237 @@
...
+    """ctx.inputs correctly accepts dynamic attribute assignments via setattr().
+
+    This verifies the behavior required by the @nominal.input system,
+    which injects data post-instantiation via setattr(ctx.inputs, name, value).
+    """
+    ctx.inputs.dataset = Path("/data/bags/run1.bag")
+    ctx.inputs.my_folder = Path("/data/my_folder")
+
+    assert ctx.inputs.dataset == Path("/data/bags/run1.bag")
+    assert ctx.inputs.my_folder == Path("/data/my_folder")
```
xaru8145: Let's update this test to use setattr(ctx.inputs, 'dataset', ...) as suggested in the previous review instead of direct assignment. Since the benchmark engine discovers `@nominal.input` hook names as strings at runtime, it will rely on `setattr()` to inject them into the namespace. Testing with this specific mechanism ensures `ctx.inputs` is behaving as the 'receptive container' our dynamic hook system requires.

src/lambkin/core/ctx/context.py
teresa-ortega: Implemented in 4221a85

test/core/test_context.py
teresa-ortega: Implemented in 4221a85

teresa-ortega:
> Some changes are required. And, I have some questions:
> 
>     * What happens if we provide the same key for 2 variations e.g. `sensor_model` twice? Would it throw an exception?
> 
>     * What happens if we provide the same values for a given key? For example,  `num_particles: [10, 10]`?

This will be addressed in a future PR regarding named_product to account for these cases

test/core/test_source.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def test_source_repr():
```
xaru8145, line 46:
```suggestion
def test_source_repr():
    """Source.__repr__ returns a human-readable string."""
    test_path = Path("/home/user/benchmarks/test.py")
    source = Source(path=test_path)
    assert repr(source) == f"Source( path = {test_path})"
```

src/lambkin/core/ctx/source.py
```diff
@@ -0,0 +1,45 @@
...
+class Source:
+    """Describes the benchmark source file.
+
+    Attributes:
+    ----------
+    path:
+        Path to the source file of the benchmark.
+    """
+
+    def __init__(self, path: Path) -> None:
```
xaru8145:
```suggestion
    def __init__(self, path: Path | str) -> None:
```

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,186 @@
...
+
+
+def _variation_folder_name(index: int) -> str:
+    """Build the variation folder name from its index."""
+    return f"var_{index + 1}"
+
+
+def _iteration_folder_name(iteration: int) -> str:
+    """Build the per-iteration folder name."""
+    return f"iter_{iteration}"
```
xaru8145: Why are we adding +1 in variation but not in iteration? Just curious

src/lambkin/core/ctx/context.py
```diff
@@ -18,3 +18,186 @@
...
+    This means a bag/ folder is only created if the user accesses
+    ctx.output.bag_dir, and same for metrics/ and any other subfolder.
+
+    Attributes:
+    ----------
+    variation_dir:
+        Root folder for this variation (e.g. <source.path.parent>/var_1/).
+    iteration_dir:
+        Folder for the current iteration (e.g. <source.path.parent>/var_1/iter_0/).
+    bag_dir:
```
xaru8145: From what I recall in our last conversation, we decided it would be better to have all bag and metrics inside the iteration folder as to simplify our lookup and avoid adding logic to conditionally look for data in nested folders.

Also, I believe that having all variation folders in source.path.parent can get messy very quickly. I would suggest adding a `results` folder in between (like before) or `case` to keep the multiple variations folders from cluttering the source path directory. What do you think @teresa-ortega?

xaru8145 (changes requested):
Almost there.

test/core/test_source.py
teresa-ortega: Implemented in c4bd6c8

test/core/test_source.py
teresa-ortega: Solved in 12a27a2

test/core/test_source.py
teresa-ortega: Solved in 12a27a2

test/core/test_source.py
teresa-ortega: Solved in 12a27a2

test/core/test_source.py
teresa-ortega: Solved in 12a27a2

src/lambkin/core/ctx/context.py
teresa-ortega: You're totally right, this is a real issue. We need the +1 because having var_0 or iter_0 doesn't make sense.
Fixed in 61b7905

xaru8145: This is an implementation choice. Starting at 1 is more human readable, starting at 0 is more a developer experience. There is no right or wrong here, we just need to make a choice. What's relevant is to have both iterations and variations be consistent and follow the same logic, either one or the other.

src/lambkin/core/ctx/context.py
teresa-ortega: I completely agree, I've been thinking about this too. Keeping everything directly inside the iteration folder is the best way to avoid issues with the folder flow later on. Furthermore, since one of the project's main goals is to have all the folder logic so the user doesn't have to input paths manually, so I think to create a results folder inside source.path.parent makes perfect sense

test/core/test_context.py
teresa-ortega: Solved in 9d536de

src/lambkin/core/ctx/source.py
teresa-ortega: Solved in 9d536de

src/lambkin/core/ctx/context.py
teresa-ortega: I implemented in 9d536de

src/lambkin/core/ctx/context.py
teresa-ortega: I completely agree about keeping them consistent. Since we want the tool to be as user-friendly as possible, I vote for the human-readable approach: starting both iterations and variations at 1. I'll make sure they both follow this exact logic

test/core/test_context.py
xaru8145: Not solved in this commit

teresa-ortega: Solved 2f84a12

xaru8145 (approved):
LGTM

PR #136, opened 2026-03-20
Phase 3: rename variation to variant in Context class

teresa-ortega:
### Proposed changes

Renamed the variantions attribute to variant within the Context class. This change was made to align our naming conventions with the expected base SDK.

#### Type of change

### Checklist

### Additional comments

### DIFF
--- a/src/lambkin/core/ctx/context.py
+++ b/src/lambkin/core/ctx/context.py
@@ -94,10 +94,10 @@ class Context:
         self,
         variant: dict[str, Any],
         iteration: int,
-        output_dir: Path | str,
         options: dict[str, Any],
         source: Source,
         variant_index: int = 0,
+        output_dir: Path | str | None = None,
     ) -> None:
         """Initialize a Context for one (variant, iteration) benchmark run.
 
--- a/src/lambkin/core/ctx/context.py
+++ b/src/lambkin/core/ctx/context.py
@@ -102,17 +102,17 @@ class Context:
         """Initialize a Context for one (variant, iteration) benchmark run.
 
         Only the parameters needed to build the namespaced sub-objects are
-        stored at construction time. Output subfolders for variation and iteration
+        stored at construction time. Output subfolders for variant and iteration
         are created immediately, while any other subfolder is created on first access.
 
         Parameters
         ----------
         variant : dict
             Algorithm parameters for this run, as defined by the user.
-            All key-value pairs are exposed as attributes on "ctx.variation".
+            All key-value pairs are exposed as attributes on "ctx.variant".
         iteration : int
-            Zero-based repetition index within this variation.
-            Controls the "iter_<N>" subfolder name under the variation directory.
+            Zero-based repetition index within this variant.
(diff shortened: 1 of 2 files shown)


xaru8145 (changes requested):
@teresa-ortega I see 2 separate fixes in this PR: one for the variation and another for the optionality of path. Let's keep them as separate PRs for traceability and separation of concerns. 

Also, add the phase in brackets in the PR title like done so far. And better refer to the class `Context` rather than the file itself.

In terms of the path optionality, I suggest adding a test case where you construct the object without providing the output to assert the behavior is the expected (but in a separate PR as noted above).

src/lambkin/core/ctx/context.py
```diff
@@ -92,22 +92,22 @@ class Context:
...
         options: dict[str, Any],
         source: Source,
-        variation_index: int = 0,
+        variant_index: int = 0,
     ) -> None:
-        """Initialize a Context for one (variation, iteration) benchmark run.
+        """Initialize a Context for one (variant, iteration) benchmark run.
 
         Only the parameters needed to build the namespaced sub-objects are
         stored at construction time. Output subfolders for variation and iteration
```
```suggestion
        stored at construction time. Output subfolders for variant and iteration
```

src/lambkin/core/ctx/context.py
```diff
@@ -92,22 +92,22 @@ class Context:
...
         are created immediately, while any other subfolder is created on first access.
 
         Parameters
         ----------
-        variation : dict
+        variant : dict
             Algorithm parameters for this run, as defined by the user.
             All key-value pairs are exposed as attributes on "ctx.variation".
```
```suggestion
            All key-value pairs are exposed as attributes on "ctx.variant".
```

src/lambkin/core/ctx/context.py
```suggestion
            Zero-based repetition index within this variant.
```

```suggestion
            Controls the "iter_<N>" subfolder name under the variant directory.
```

xaru8145 (changes requested):
Still see some references in the docu to variation.

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in 5694f0b

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in 5694f0b

src/lambkin/core/ctx/context.py
teresa-ortega: Solved in 5694f0b

teresa-ortega: Solved in 5694f0b

xaru8145: Still not solved: https://github.com/Ekumen-OS/lambkin/blob/5694f0b6812b0f618124189d6df5f2c5a4c521ee/src/lambkin/core/ctx/context.py#L124

xaru8145: @teresa-ortega PTAL

teresa-ortega: Solved the comment in 96575ab

xaru8145 (approved):
LGTM

PR #137, opened 2026-03-20
Phase 3: make output_dir optional in Context class

teresa-ortega:
### Proposed changes

This PR makes output_dir an optional parameter in Context class.

When not provided, output_dir defaults to source.path.parent / "results",so the benchmark results are written next to the benchmark script without requiring the user to specify a path explicitly.

A test is added to verify this default behaviour — it constructs a Context without passing output_dir and asserts that the output directory is resolved relative to the benchmark script location.

#### Type of change

### Checklist

### Additional comments

### DIFF
--- a/test/core/test_context.py
+++ b/test/core/test_context.py
@@ -215,14 +215,14 @@ def test_context_default_output_dir(tmp_path):
     """Default output_dir resolves relative to the benchmark script.
 
     When output_dir is not provided, it defaults to source.path.parent /
-    OutputInfo.BENCHMARKS_DIRNAME.
+    Context.BENCHMARKS_DIRNAME.
     """
     script = tmp_path / "my_benchmark.py"
     script.touch()
 
     source = Source(path=script)
     ctx = Context(
-        variation={"sensor_model": "beam", "num_particles": 10},
+        variant={"sensor_model": "beam", "num_particles": 10},
         iteration=0,
         options={},
         source=source,
--- a/test/core/test_context.py
+++ b/test/core/test_context.py
@@ -212,9 +212,10 @@ def test_iteration_stored(tmp_path, base_variant, base_options, base_source, ite
 
 
 def test_context_default_output_dir(tmp_path):
-    """When output_dir is not provided.
+    """Default output_dir resolves relative to the benchmark script.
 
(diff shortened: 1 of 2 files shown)


test/core/test_context.py
```diff
@@ -211,3 +211,21 @@ def test_iteration_stored(
...
+    script.touch()
+
+    source = Source(path=script)
+    ctx = Context(
+        variation={"sensor_model": "beam", "num_particles": 10},
+        iteration=0,
+        options={},
+        source=source,
+    )
+    assert ctx.output.variation_dir.parent == tmp_path / "results"
```
xaru8145: Hardcoding strings like "results" in your tests can be problematic because if you ever decide to change the default folder name to "benchmarks_out" or "data", you’ll have to hunt down and fix tests.

One option is to define it as a class constant. Another option is to modify OutputInfo to store the base_dir so you can later check it in the test. There might be other options, but you get the idea.

xaru8145 (changes requested):
Tests now show us a limitation of the current implementation so I advise in favor of taking the opportunity for a bit of refactor.

Also, I think this is a fix, not a feature as explained in the PR description.

test/core/test_context.py
teresa-ortega: Solved in 77f49ab

test/core/test_context.py
xaru8145: Review documentation. 
- Why adding a dot after "provided"? It does not read natural.
- Still references `OutputInfo.BENCHMARKS_DIRNAME`.

xaru8145 (changes requested):
Minor change to go.

Just a comment from the top of my mind. Right now, `OutputInfo` is just a placeholder/struct for 3 paths with no logic whatsoever. Maybe it makes sense to move the path construction logic from `Context` into `OutputInfo` itself, giving it a proper constructor that takes `base_dir`, `variation_index`, and `iteration` and derives `variation_dir` and `iteration_dir` internally. This would make `OutputInfo` a cohesive object that owns everything path-related, and reduce `Context construction to just deciding the `base_dir` and delegating the rest. 

No need to do now, but let's keep it in mind and maybe have a discussion about it.

test/core/test_context.py
teresa-ortega: Solved in 475c26b

xaru8145 (approved):
LGTM

PR #138, opened 2026-03-20
Phase 3: core/decorators/benchmark

teresa-ortega:
### Proposed changes

Adds the `benchmark` decorator, which orchestrates the benchmark execution loop. It parses CLI options registered by @lambkin.option once before the loop using an internal click parser, then creates a fresh Context for each (variant, iteration) pair, injecting the parsed options, the source inferred from the script path, and the variant index.

CLI options are parsed once and reused across all iterations to avoid redundant parsing.

Relates to:
- #136 
- #137

#### Type of change

💥 No breaking changes.

### Checklist

### Additional comments

N/A

### DIFF
--- a/src/lambkin/__init__.py
+++ b/src/lambkin/__init__.py
@@ -13,3 +13,11 @@
 # limitations under the License.
 
 """LAMBKIN: Localization And Mapping BenchmarKINg SDK."""
+
+from lambkin.core.decorators.benchmark import benchmark
+from lambkin.core.decorators.option import option
+
+__all__ = [
+    "option",
+    "benchmark",
+]
--- a/src/lambkin/__init__.py
+++ b/src/lambkin/__init__.py
@@ -14,10 +14,12 @@
 
 """LAMBKIN: Localization And Mapping BenchmarKINg SDK."""
 
+from lambkin.common.named_product import named_product
 from lambkin.core.decorators.benchmark import benchmark
 from lambkin.core.decorators.option import option
 
 __all__ = [
     "option",
     "benchmark",
+    "named_product",
 ]
--- a/src/lambkin/__init__.py
+++ b/src/lambkin/__init__.py
@@ -14,12 +14,11 @@
 
 """LAMBKIN: Localization And Mapping BenchmarKINg SDK."""
 
-from lambkin.common.named_product import named_product
-from lambkin.core.decorators.benchmark import benchmark
-from lambkin.core.decorators.option import option
+from lambkin.common import named_product
+from lambkin.core.decorators import benchmark, option
 
 __all__ = [
-    "option",
     "benchmark",
     "named_product",
+    "option",
 ]
--- a/src/lambkin/core/decorators/__init__.py
+++ b/src/lambkin/core/decorators/__init__.py
@@ -21,5 +21,5 @@ Provides the decorators used to define and configure benchmark functions:
 
 - @lambkin.benchmark : It collects CLI option definitions registered by @option,
   parses them once before the loop using an internal click parser, and injects
-  the resulting values into a Context class on each (variation, iteration)
+  the resulting values into a Context class on each (variant, iteration)
 """
--- a/src/lambkin/core/decorators/__init__.py
+++ b/src/lambkin/core/decorators/__init__.py
@@ -18,12 +18,8 @@ Provides the decorators used to define and configure benchmark functions:
 
 - @lambkin.option : registers CLI options so the @benchmark decorator can inject
   them into "ctx.options".
+
+- @lambkin.benchmark : It collects CLI option definitions registered by @option,
+  parses them once before the loop using an internal click parser, and injects
+  the resulting values into a Context class on each (variation, iteration)
 """
-
-# from .option import option
-# from .benchmark import benchmark
-
-# __all__ = [
-#     "option",
-#     "benchmark",
-# ]
--- a/src/lambkin/core/decorators/__init__.py
+++ b/src/lambkin/core/decorators/__init__.py
@@ -25,9 +25,9 @@ Provides the decorators used to define and configure benchmark functions:
 """
 
 from .benchmark import benchmark
-from .option import Option
+from .option import option
 
 __all__ = [
-    "Option",
+    "option",
     "benchmark",
 ]
--- a/src/lambkin/core/decorators/__init__.py
+++ b/src/lambkin/core/decorators/__init__.py
@@ -23,3 +23,11 @@ Provides the decorators used to define and configure benchmark functions:
   parses them once before the loop using an internal click parser, and injects
   the resulting values into a Context class on each (variant, iteration)
 """
+
+from .benchmark import benchmark
+from .option import Option
+
+__all__ = [
+    "Option",
+    "benchmark",
+]
--- a/src/lambkin/core/decorators/__init__.py
+++ b/src/lambkin/core/decorators/__init__.py
@@ -20,8 +20,10 @@ Provides the decorators used to define and configure benchmark functions:
   them into "ctx.options".
 """
 
-from .option import option
+# from .option import option
+# from .benchmark import benchmark
 
-__all__ = [
-    "option",
-]
+# __all__ = [
+#     "option",
+#     "benchmark",
+# ]
--- a/src/lambkin/core/decorators/__init__.py
+++ b/src/lambkin/core/decorators/__init__.py
@@ -16,18 +16,18 @@
 
 Provides the decorators used to define and configure benchmark functions:
 
-- @lambkin.option : registers CLI options so the @benchmark decorator can inject
-  them into "ctx.options".
-
 - @lambkin.benchmark : It collects CLI option definitions registered by @option,
   parses them once before the loop using an internal click parser, and injects
   the resulting values into a Context class on each (variant, iteration)
+
+- @lambkin.option : registers CLI options so the @benchmark decorator can inject
+  them into "ctx.options".
 """
 
 from .benchmark import benchmark
 from .option import option
 
 __all__ = [
-    "option",
     "benchmark",
+    "option",
 ]
--- a/src/lambkin/core/decorators/benchmark.py
+++ b/src/lambkin/core/decorators/benchmark.py
@@ -14,6 +14,7 @@
 
 """Benchmark loop decorator for lambkin."""
 
+import inspect
 import sys
 
 import click
@@ -22,13 +23,13 @@ from lambkin.core.ctx.context import Context
 from lambkin.core.ctx.source import Source
 
 
-def _parse_options(fn):
+def _parse_options(fn, cli_args):
     """Parse CLI options from fn.__lambkin_options__ and return a dict."""
     registered = getattr(fn, "__lambkin_options__", [])
     if not registered:
         return {}
     cmd = click.Command(name="benchmark", params=registered)
-    click_ctx = cmd.make_context("benchmark", sys.argv[1:], allow_extra_args=True)
+    click_ctx = cmd.make_context("benchmark", list(cli_args))
     return click_ctx.params
 
 
@@ -50,9 +51,10 @@ def benchmark(variants, num_iterations):
     """
 
     def decorator(fn):
-        def wrapper():
-            options = _parse_options(fn)
-            source = Source(path=fn.__code__.co_filename)
+        def wrapper(args=None):
+            cli_args = sys.argv[1:] if args is None else args
+            options = _parse_options(fn, cli_args)
+            source = Source(path=inspect.getfile(fn))
             for variation_index, variation in enumerate(variants):
                 for iteration in range(num_iterations):
                     ctx = Context(
--- a/src/lambkin/core/decorators/benchmark.py
+++ b/src/lambkin/core/decorators/benchmark.py
@@ -12,7 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Benchmark loop decorator for lambkin."""
+"""Benchmark loop decorator for lambkin.
+
+Provides the @benchmark decorator, which drives the execution loopover all
+variants and iterations. It collects CLI option definitions registered by
+@option, parses them once before the loop using an internal click parser, and
+injects the resulting values into a Context class on each (variation, iteration)
+pair.
+
+"""
 
 import inspect
 import sys
--- a/src/lambkin/core/decorators/benchmark.py
+++ b/src/lambkin/core/decorators/benchmark.py
@@ -66,9 +66,13 @@ def benchmark(variants, num_iterations):
             for variation_index, variation in enumerate(variants):
                 for iteration in range(num_iterations):
                     ctx = Context(
+                        # TODO(teresa-ortega): change the name variation.
+                        # variation-> variant.
                         variation=variation,
                         iteration=iteration,
                         output_dir=".",
+                        # TODO(teresa-ortega): change the output_dir
+                        # implementation.
                         options=options,
                         source=source,
                         variation_index=variation_index,
--- a/src/lambkin/core/decorators/benchmark.py
+++ b/src/lambkin/core/decorators/benchmark.py
@@ -17,7 +17,7 @@
 Provides the @benchmark decorator, which drives the execution loopover all
 variants and iterations. It collects CLI option definitions registered by
 @option, parses them once before the loop using an internal click parser, and
-injects the resulting values into a Context class on each (variation, iteration)
+injects the resulting values into a Context class on each (variant, iteration)
 pair.
 
 """
@@ -46,17 +46,17 @@ def benchmark(variants, num_iterations):
     """Drive the benchmark execution loop over all variants and iterations.
 
     Parses CLI options registered by @lambkin.option once before the loop,
-    then creates a Context for each (variants, iteration) pair and calls
+    then creates a Context for each (variant, iteration) pair and calls
     the decorated function with it.
 
     Parameters
     ----------
     variants : iterable of dict
-        Sequence of variation dicts to sweep over. Each dict is exposed
-        as attributes on ctx.variation.
+        Sequence of variant dicts to sweep over. Each dict is exposed
+        as attributes on ctx.variant.
     num_iterations : int
-        Number of times to repeat each variation. Controls the iter_<N>
-        subfolders under each variation directory.
+        Number of times to repeat each variant. Controls the iter_<N>
+        subfolders under each variant directory.
     """
 
     def decorator(fn):
--- a/src/lambkin/core/decorators/benchmark.py
+++ b/src/lambkin/core/decorators/benchmark.py
@@ -22,6 +22,7 @@ pair.
 
 """
 
+import functools
 import inspect
 import sys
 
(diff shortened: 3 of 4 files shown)


src/lambkin/core/decorators/__init__.py
xaru8145, line 1: Why is everything in this file commented?

src/lambkin/core/decorators/benchmark.py
```diff
@@ -13,3 +13,58 @@
...
+    def decorator(fn):
+        def wrapper():
+            options = _parse_options(fn)
+            source = Source(path=fn.__code__.co_filename)
+            for variation_index, variation in enumerate(variants):
+                for iteration in range(num_iterations):
+                    ctx = Context(
+                        variation=variation,
+                        iteration=iteration,
+                        output_dir=".",
```
xaru8145: Hardcoding `"."` means results land wherever the process is invoked from, not relative to the benchmark script. 

The `Context` constructor already handles a `None` fallback to `source.path.parent / "results"`, so passing None (or simply omitting a hardcoded ".") and letting Context do its job would be more correct.

test/core/test_benchmark.py
xaru8145, line 1: The test file tests `_parse_options` only indirectly via `test_benchmark_options_defaults_injected`. A direct unit test for it (with a mock function carrying `__lambkin_options__`) would be more robust.

src/lambkin/core/decorators/benchmark.py
```diff
@@ -13,3 +13,58 @@
...
+        as attributes on ctx.variation.
+    num_iterations : int
+        Number of times to repeat each variation. Controls the iter_<N>
+        subfolders under each variation directory.
+    """
+
+    def decorator(fn):
+        def wrapper():
+            options = _parse_options(fn)
+            source = Source(path=fn.__code__.co_filename)
```
xaru8145: `fn.__code__.co_filename` can return a relative path depending on how Python was invoked, and it points to the wrong file if the function is defined in a compiled/frozen context. Instead, use `inspect.getfile(func)`, which is the industry standard for reliable source discovery.

src/lambkin/core/decorators/benchmark.py
```diff
@@ -13,3 +13,58 @@
 # limitations under the License.
 
 """Benchmark loop decorator for lambkin."""
```
xaru8145: Make the documentation more descriptive and detailed, like for the other decorators and modules.

src/lambkin/core/decorators/benchmark.py
```diff
@@ -13,3 +13,58 @@
...
+
+
+def _parse_options(fn):
+    """Parse CLI options from fn.__lambkin_options__ and return a dict."""
+    registered = getattr(fn, "__lambkin_options__", [])
+    if not registered:
+        return {}
+    cmd = click.Command(name="benchmark", params=registered)
+    click_ctx = cmd.make_context("benchmark", sys.argv[1:], allow_extra_args=True)
+    return click_ctx.params
```
xaru8145: Calling `sys.argv[1:]` here means any pytest arguments (-v, -k, etc.) get fed into Click's parser. The `allow_extra_args=True` flag mitigates crashes but silently swallows test runner flags as benchmark options. Therefore, `_parse_options` should accept an explicit args list.

```suggestion
def _parse_options(fn, cli_args):
    """Parse CLI options from fn.__lambkin_options__ and return a dict."""
    registered = getattr(fn, "__lambkin_options__", [])
    if not registered:
        return {}
    cmd = click.Command(name="benchmark", params=registered)
    click_ctx = cmd.make_context("benchmark", list(cli_args))
    return click_ctx.params
```

test/core/test_benchmark.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @benchmark(variants=variants, num_iterations=1)
+    @option("--clock-rate", default=100.0)
+    @option("--sensor-topic", default="/scan")
+    def fn(ctx):
+        contexts.append(ctx)
+
+    fn()
+    assert contexts[0].options.clock_rate == 100.0
+    assert contexts[0].options.sensor_topic == "/scan"
```
xaru8145: The test `test_benchmark_options_defaults_injected` passes today only because it relies on defaults, but it would fail if any unrecognized pytest flag slipped through as a Click error.

It's worth adding a test that verifies overrides by injection, which is allowed with the suggested code reviews done in the benchmark decorator.

src/lambkin/core/decorators/benchmark.py
```diff
@@ -13,3 +13,58 @@
...
+        Sequence of variation dicts to sweep over. Each dict is exposed
+        as attributes on ctx.variation.
+    num_iterations : int
+        Number of times to repeat each variation. Controls the iter_<N>
+        subfolders under each variation directory.
+    """
+
+    def decorator(fn):
+        def wrapper():
+            options = _parse_options(fn)
```
xaru8145: Related to the comment above in `_parse_options` function, the standard pattern is to let the `wrapper()` accept `args=None` and default to `sys.argv[1:]`, making it injectable for tests.

```suggestion
    def decorator(fn):
        def wrapper(args=None):
            cli_args = sys.argv[1:] if args is None else args
            options = _parse_options(fn, cli_args)
```

src/lambkin/core/decorators/benchmark.py
```diff
@@ -13,3 +13,58 @@
...
+    variants : iterable of dict
+        Sequence of variation dicts to sweep over. Each dict is exposed
+        as attributes on ctx.variation.
+    num_iterations : int
+        Number of times to repeat each variation. Controls the iter_<N>
+        subfolders under each variation directory.
+    """
+
+    def decorator(fn):
+        def wrapper():
```
xaru8145: We need to make the decorator 'transparent.' By adding `@functools.wraps(fn)` to the wrapper, we copy the original function's name and docstring over. This ensures that when a user runs `help(your_benchmark)`, they see their own documentation instead of the SDK's internal wrapper logic.

xaru8145 (changes requested):

src/lambkin/core/decorators/benchmark.py
teresa-ortega: I completely agree with you. That's just a temporary implementation since the output_dir PR hasn't been merged yet

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in 62f4b25

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in 552ddd2

src/lambkin/core/decorators/benchmark.py
xaru8145: In this case, I suggest you do 2 things:
- List in the PR description that there is a related PR and add a link.
- Add a TODO comment so the reviewer knows you will address this and serves also as a note to future self so you don't forget.

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in 552ddd2

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in 62f4b25

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in 662aace

src/lambkin/core/decorators/__init__.py
teresa-ortega: I delete the comment lines in 632f610

test/core/test_benchmark.py
teresa-ortega: Solved in ff1bb03

test/core/test_benchmark.py
teresa-ortega: Implemented in ff1bb03

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in 6aca587

src/lambkin/core/decorators/benchmark.py
```diff
@@ -12,4 +12,73 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Benchmark loop decorator for lambkin."""
+"""Benchmark loop decorator for lambkin.
+
+Provides the @benchmark decorator, which drives the execution loopover all
+variants and iterations. It collects CLI option definitions registered by
+@option, parses them once before the loop using an internal click parser, and
+injects the resulting values into a Context class on each (variation, iteration)
```
xaru8145:
```suggestion
injects the resulting values into a Context class on each (variant, iteration)
```

src/lambkin/core/decorators/__init__.py
```diff
@@ -18,10 +18,8 @@
...
-"""
-
-from .option import option
 
-__all__ = [
-    "option",
-]
+- @lambkin.benchmark : It collects CLI option definitions registered by @option,
+  parses them once before the loop using an internal click parser, and injects
+  the resulting values into a Context class on each (variation, iteration)
```
xaru8145:
```suggestion
  the resulting values into a Context class on each (variant, iteration)
```

src/lambkin/core/decorators/benchmark.py
```diff
@@ -12,4 +12,73 @@
...
+    the decorated function with it.
+
+    Parameters
+    ----------
+    variants : iterable of dict
+        Sequence of variation dicts to sweep over. Each dict is exposed
+        as attributes on ctx.variation.
+    num_iterations : int
+        Number of times to repeat each variation. Controls the iter_<N>
+        subfolders under each variation directory.
```
xaru8145:
```suggestion
        Sequence of variant dicts to sweep over. Each dict is exposed
        as attributes on ctx.variant.
    num_iterations : int
        Number of times to repeat each variant. Controls the iter_<N>
        subfolders under each variant directory.
```

src/lambkin/core/decorators/benchmark.py
```diff
@@ -12,4 +12,73 @@
...
+    cmd = click.Command(name="benchmark", params=registered)
+    click_ctx = cmd.make_context("benchmark", list(cli_args))
+    return click_ctx.params
+
+
+def benchmark(variants, num_iterations):
+    """Drive the benchmark execution loop over all variants and iterations.
+
+    Parses CLI options registered by @lambkin.option once before the loop,
+    then creates a Context for each (variants, iteration) pair and calls
```
xaru8145:
```suggestion
    then creates a Context for each (variant, iteration) pair and calls
```

test/core/test_benchmark.py
xaru8145, line 1: - Add a metadata introspection test. Example:
```py
def test_benchmark_preserves_metadata(variants):
    """Verify that @benchmark preserves the decorated function's name and docstring."""
    @benchmark(variants=variants, num_iterations=1)
    def my_cool_benchmark(ctx):
        """Standard docstring."""
        pass

    assert my_cool_benchmark.__name__ == "my_cool_benchmark"
    assert my_cool_benchmark.__doc__ == "Standard docstring."
```

- Every test in that file that calls the decorated `fn()` is currently polluting your environment by "silently" creating folders in your project root every time it runs. We need a way to inject the output_dir from the test so we can point to a temp directory. See other code comments. 
 
- Add a test to verify the base file is as expected, proving te Orchestrator uses inspect to find the file, passes output_dir=None to the Context, and the Context uses the Source to build the path. To avoid polluting the environment as mentioned above you can use `[mocks](https://docs.python.org/3/library/unittest.mock.html)` to inject a Source that points to a temporary directory.

src/lambkin/core/decorators/benchmark.py
```diff
@@ -12,4 +12,73 @@
...
+        def wrapper(args=None):
+            cli_args = sys.argv[1:] if args is None else args
+            options = _parse_options(fn, cli_args)
+            source = Source(path=inspect.getfile(fn))
+            for variant_index, variant in enumerate(variants):
+                for iteration in range(num_iterations):
+                    ctx = Context(
+                        variant=variant,
+                        iteration=iteration,
+                        output_dir=".",
```
xaru8145:
```suggestion
                        output_dir=output_dir,
```

src/lambkin/core/decorators/benchmark.py
```diff
@@ -12,4 +12,73 @@
...
+        Sequence of variation dicts to sweep over. Each dict is exposed
+        as attributes on ctx.variation.
+    num_iterations : int
+        Number of times to repeat each variation. Controls the iter_<N>
+        subfolders under each variation directory.
+    """
+
+    def decorator(fn):
+        @functools.wraps(fn)
+        def wrapper(args=None):
```
xaru8145: In this way we can inject `output_dir` from the tests so that we do not create directories that pollute the workspace.
```suggestion
        def wrapper(args=None, output_dir=None):
```

src/lambkin/core/decorators/__init__.py
xaru8145, line 1: This file is empty now? Why not exposing the decorators here?

xaru8145 (changes requested):

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in eda77c7

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in eda77c7

src/lambkin/core/decorators/__init__.py
teresa-ortega: Solved in eda77c7

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in eda77c7

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in 4daf9f0

src/lambkin/core/decorators/benchmark.py
teresa-ortega: implementation(benchmark.py) in 4daf9f0

teresa-ortega: Test in b364198

test/core/test_benchmark.py
teresa-ortega: Solved in b364198

src/lambkin/core/decorators/__init__.py
teresa-ortega: You are right, I implemented in 8cb46ea

src/lambkin/__init__.py
```diff
@@ -13,3 +13,13 @@
 # limitations under the License.
 
 """LAMBKIN: Localization And Mapping BenchmarKINg SDK."""
+
+from lambkin.common.named_product import named_product
+from lambkin.core.decorators.benchmark import benchmark
+from lambkin.core.decorators.option import option
```
xaru8145: `lambkin/__init__.py` is not using `core/decorators/__init__.py `, it imports directly from the module files instead.

```suggestion
from lambkin.common import named_product
from lambkin.core.decorators import benchmark, option
```

src/lambkin/__init__.py
```diff
@@ -13,3 +13,13 @@
...
+
+__all__ = [
+    "option",
+    "benchmark",
+    "named_product",
+]
```
xaru8145, line 24: Sort alphabetically (standard).

src/lambkin/core/decorators/__init__.py
```diff
@@ -18,10 +18,16 @@
...
+  parses them once before the loop using an internal click parser, and injects
+  the resulting values into a Context class on each (variant, iteration)
 """
 
+from .benchmark import benchmark
 from .option import option
 
 __all__ = [
     "option",
+    "benchmark",
```
xaru8145: Sort alphabetically. Update also the docu order according to it.

src/lambkin/core/decorators/benchmark.py
```diff
@@ -12,4 +12,71 @@
...
+            source = Source(path=inspect.getfile(fn))
+            for variant_index, variant in enumerate(variants):
+                for iteration in range(num_iterations):
+                    ctx = Context(
+                        variant=variant,
+                        iteration=iteration,
+                        output_dir=output_dir,
+                        options=options,
+                        source=source,
+                        variant_index=variant_index,
```
xaru8145, line 85: Re-order to match the order from the Context constructor:

```suggestion
                        variant=variant,
                        iteration=iteration,
                        options=options,
                        source=source,
                        variant_index=variant_index,
                        output_dir=output_dir,
```

test/core/test_benchmark.py
xaru8145, line 1: There is no test covering what happens when `variants` is an empty list. `wrapper()` would silently do nothing, no iterations, no error. Whether that should raise or just be a no-op is a design decision, but it's worth a test either way to pin down the intended behavior, especially since `named_product` has a related TODO about empty lists raising an exception.

xaru8145 (changes requested):
Minor things to go. Good job!

src/lambkin/__init__.py
teresa-ortega: Solved in 927de2f

src/lambkin/__init__.py
teresa-ortega: Solved in 927de2f

src/lambkin/core/decorators/__init__.py
teresa-ortega: Solved in 40224b0

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in 3a3cb9e

test/core/test_benchmark.py
teresa-ortega: I agree that a silent no-op can be confusing for the user. I've decided to take a 'fail fast' approach: if the variants list is empty, the decorator will now explicitly raise a ValueError. I've also added the corresponding test to pin down this behavior.

Regarding the named_product TODO, I will resolve it in a future PR by raising a different exception for those particular cases.

Implemented in b66fcbe

xaru8145: Update the documentation of the decorator to explain when it throws e.g.

```py
Raises
------
ValueError
    If variants is an empty list.
```

teresa-ortega: Solved in a8c199f

src/lambkin/core/decorators/benchmark.py
```diff
@@ -19,7 +19,7 @@
 @option, parses them once before the loop using an internal click parser, and
 injects the resulting values into a Context class on each (variant, iteration)
 pair.
-
+Raises ValueError if variants is empty.
```
xaru8145, line 22: Not needed here. This just describes the general behavior. Just in the decorator.

xaru8145 (approved):
LGTM! Minor comment.

PR #139, opened 2026-03-26
Phase 3: core/decorators/input

teresa-ortega:
### Proposed changes

This PR introduces a dynamic input registration system for benchmarks using a functional hook pattern. To keep the core benchmark decorator clean and maintain a strict separation of concerns, the registration logic was extracted into a dedicated InputRegistry. The decorated benchmark function now dynamically exposes an .input method, allowing users to register input hooks. 
Additionally, the benchmark context was updated. At runtime, right before the main benchmark executes, these registered hooks are evaluated with the actual context, and their return values are automatically mapped to the context inputs based on the function names.

#### Type of change

💥 No breaking changes.

### Checklist

### Additional comments

N/A

### DIFF
--- a/src/lambkin/core/decorators/benchmark.py
+++ b/src/lambkin/core/decorators/benchmark.py
@@ -30,6 +30,7 @@ import click
 
 from lambkin.core.ctx.context import Context
 from lambkin.core.ctx.source import Source
+from lambkin.core.decorators.input import InputRegistry
 
 
 def _parse_options(fn, cli_args):
@@ -70,6 +71,8 @@ def benchmark(variants, num_iterations):
         )
 
     def decorator(fn):
+        inputs = InputRegistry()
+
         @functools.wraps(fn)
         def wrapper(args=None, output_dir=None):
             cli_args = sys.argv[1:] if args is None else args
@@ -85,8 +88,10 @@ def benchmark(variants, num_iterations):
                         variant_index=variant_index,
                         output_dir=output_dir,
                     )
+                    inputs.resolve(ctx)
                     fn(ctx)
 
+        wrapper.input = inputs.register
         return wrapper
 
     return decorator
--- a/src/lambkin/core/decorators/benchmark.py
+++ b/src/lambkin/core/decorators/benchmark.py
@@ -19,6 +19,11 @@ variants and iterations. It collects CLI option definitions registered by
 @option, parses them once before the loop using an internal click parser, and
 injects the resulting values into a Context class on each (variant, iteration)
 pair.
+
+Input hooks registered via "@nominal.input" are managed by "InputRegistry"
+instance and resolved before the benchmark function runs on each iteration,
+injecting their return values into "ctx.inputs".
+
 Raises ValueError if variants is empty.
 """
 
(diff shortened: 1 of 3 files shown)


src/lambkin/core/decorators/input.py
```diff
@@ -12,4 +12,33 @@
...
+        """Initialize the InputRegistry."""
+        self._hooks = []
+
+    def register(self, hook_fn):
+        """Decorator used to register a function as an input provider."""
+        self._hooks.append(hook_fn)
+        return hook_fn
+
+    def resolve(self, ctx):
+        """Resolve all registered input hooks.
```
xaru8145: A hook that forgets to return, or a DVC call that returns None, poisons `ctx.inputs` silently. Add a guard e.g. raise `ValueError`.

src/lambkin/core/decorators/benchmark.py
```diff
@@ -85,8 +94,10 @@ def wrapper(args=None, output_dir=None):
                         variant_index=variant_index,
                         output_dir=output_dir,
                     )
+                    inputs.resolve(ctx)
```
xaru8145: Inputs are re-resolved on every iteration. I know it needs the context but let's see if we can find a more efficient way to deal with this.

src/lambkin/core/decorators/input.py
```diff
@@ -12,4 +12,33 @@
...
+    """Manages the registration and resolution of input hooks for a benchmark."""
+
+    def __init__(self):
+        """Initialize the InputRegistry."""
+        self._hooks = []
+
+    def register(self, hook_fn):
+        """Decorator used to register a function as an input provider."""
+        self._hooks.append(hook_fn)
+        return hook_fn
```
xaru8145, line 74: What happens if 2 hooks share the same name?

xaru8145 (changes requested):
First pass.

src/lambkin/core/decorators/input.py
teresa-ortega: Implemented in c858ed6

src/lambkin/core/decorators/input.py
teresa-ortega: Solved in c858ed6

src/lambkin/core/decorators/benchmark.py
teresa-ortega: possible solution in e3bd987

src/lambkin/core/decorators/input.py
```diff
@@ -12,4 +12,58 @@
...
+            f"Hook '{hook_fn.__name__}' returned an empty or blank string."
+        )
+
+
+def _validate_hook_signature(hook_fn) -> None:
+    """Validate that the hook function accepts a single 'ctx' parameter."""
+    params = list(inspect.signature(hook_fn).parameters.keys())
+
+    if len(params) != 1:
+        raise ValueError(f"Hook '{hook_fn.__name__}' must have exactly 1 parameter, ")
```
xaru8145: Why does it end in a comma? Is it a typo or is there some text missing?

src/lambkin/core/decorators/input.py
```diff
@@ -12,4 +12,58 @@
...
+            raise ValueError(
+                f"Hook name conflict: '{hook_fn.__name__}' is already registered."
+            )
+        self._hooks.append(hook_fn)
+        return hook_fn
+
+    def resolve(self, ctx):
+        """Resolve all registered input hooks."""
+        for hook in self._hooks:
+            _validate_hook_signature(hook)
```
xaru8145: Any reason why the validation of the hook signature happens when resolving rather than when registering?

I think it could be reasonable to do it at registration time since the signature doesn't change, which would give earlier feedback.

src/lambkin/core/decorators/input.py
xaru8145, line 1: Since input hooks are resolved with a base context with dummy variant/iteration values, accessing `ctx.variant` or `ctx.iteration` inside a hook will silently receive empty/wrong values. Worth documenting this explicitly on InputRegistry and @nominal.input to avoid hard-to-debug issues.

test/core/test_input.py
xaru8145, line 1: You could add a test verifying that input hooks are called exactly once regardless of the number of variants and iterations.

xaru8145 (changes requested):
Minor things to go. The PR is ready for review, you can update the draft state.

src/lambkin/core/decorators/input.py
teresa-ortega: You are right, it is a typo error. 
Solved in 795d1b

src/lambkin/core/decorators/input.py
teresa-ortega: Solved in 762fc1

src/lambkin/core/decorators/input.py
teresa-ortega: Documented in 524b77a

test/core/test_input.py
teresa-ortega: Solved in cf56115

xaru8145 (approved):
LGTM

PR #140, opened 2026-04-07
Phase 3: core/shell/proxy

teresa-ortega:
### Proposed changes

A ShellProxy class under the shell/ subpackage as a first step towards a full shell abstraction layer. Commands like 
```python
ctx.shell.ros2.launch(...)
ctx.shell.evo_ape.bag2(...)
print [CMD]: <full command> 
```
to stdout instead of executing anything. The interface is fully generic, making it easily extensible and keeping benchmark code unchanged when swapping in a real shell implementation later.

#### Type of change

💥  No breaking changes.

### Checklist

### Additional comments
N/A

### DIFF
--- a/examples/beluga/beluga_benchmark.py
+++ b/examples/beluga/beluga_benchmark.py
@@ -24,7 +24,7 @@ import lambkin
 )
 @lambkin.option("--clock-rate", default=100.0)
 @lambkin.option("--sensor-topic", default="/scan")
-@lambkin.option("--dry-run", default=False)
+@lambkin.option("--dry-run", default=True)
 def nominal(ctx):
     """Run a nominal Beluga AMCL benchmark across sensor models and particle counts.
 
--- a/examples/beluga/beluga_benchmark.py
+++ b/examples/beluga/beluga_benchmark.py
@@ -40,7 +40,6 @@ def nominal(ctx):
     print(f"sensor_topic: {ctx.options.sensor_topic}")
     print(f"source: {ctx.source}")
     print("---")
-    ctx.shell = lambkin.ShellProxy(dry_run=ctx.options.dry_run)
     ctx.shell.ros2.launch(
         ctx.source.path.parent / "beluga.launch.xml",
         f"sensor_model:={ctx.variant.sensor_model}",
--- a/examples/beluga/beluga_benchmark.py
+++ b/examples/beluga/beluga_benchmark.py
@@ -24,6 +24,7 @@ import lambkin
 )
 @lambkin.option("--clock-rate", default=100.0)
 @lambkin.option("--sensor-topic", default="/scan")
+@lambkin.option("--dry-run", default=False)
 def nominal(ctx):
     """Run a nominal Beluga AMCL benchmark across sensor models and particle counts.
 
@@ -39,6 +40,7 @@ def nominal(ctx):
     print(f"sensor_topic: {ctx.options.sensor_topic}")
     print(f"source: {ctx.source}")
     print("---")
+    ctx.shell = lambkin.ShellProxy(dry_run=ctx.options.dry_run)
     ctx.shell.ros2.launch(
         ctx.source.path.parent / "beluga.launch.xml",
         f"sensor_model:={ctx.variant.sensor_model}",
--- a/src/lambkin/core/ctx/context.py
+++ b/src/lambkin/core/ctx/context.py
@@ -23,6 +23,8 @@ from pathlib import Path
 from types import SimpleNamespace
 from typing import Any
 
+from lambkin.core.shell import ShellProxy
+
 from .source import Source
 
 
@@ -176,7 +178,7 @@ class Context:
         # as parameters.
 
         # ctx.shell
-        # TODO(teresa-ortega): self.shell = Shell(self)
+        self.shell = ShellProxy()
 
     def _setup_directories(self) -> None:
         """Create variant and iteration directories."""
--- a/src/lambkin/core/ctx/context.py
+++ b/src/lambkin/core/ctx/context.py
@@ -178,7 +178,7 @@ class Context:
         # as parameters.
 
         # ctx.shell
-        self.shell = ShellProxy()
+        self.shell = ShellProxy(dry_run=getattr(self.options, "dry_run", False))
 
     def _setup_directories(self) -> None:
         """Create variant and iteration directories."""
--- a/src/lambkin/core/shell/proxy.py
+++ b/src/lambkin/core/shell/proxy.py
@@ -27,13 +27,14 @@ from typing import Any
 class _CommandProxy:
     """Builds a shell command lazily by chaining attribute access and calls."""
 
-    def __init__(self, parts: list[str]) -> None:
+    def __init__(self, parts: list[str], dry_run: bool = False) -> None:
         """Initialize the proxy with the command words accumulated so far."""
         self._parts = parts
+        self._dry_run = dry_run
 
     def __getattr__(self, name: str) -> _CommandProxy:
         """Append a new word to the command and return a new proxy."""
-        return _CommandProxy(self._parts + [name])
+        return _CommandProxy(self._parts + [name], self._dry_run)
 
     def __call__(self, *args: Any, **kwargs: Any) -> None:
         """Finalize and print the command.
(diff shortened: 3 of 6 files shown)


src/lambkin/core/shell/proxy.py
```diff
@@ -17,3 +17,50 @@
...
+                extra.append(flag)
+            elif value is not False:
+                extra.extend([flag, shlex.quote(str(value))])
+
+        command = " ".join(self._parts + extra)
+        print(f"[CMD]: {command}")
+
+
+class ShellProxy:
+    """Dry-run mock shell that prints commands instead of executing them."""
```
xaru8145, line 74: Worth extending the documentation with a comment explaining that this will be extended to dispatch
commands to BackgroundProcess for real execution.

src/lambkin/core/shell/proxy.py
xaru8145, line 1: Might be worth adding a `dry_run` boolean flag and just make the implementation for when it is true and leave as TODOs for when is false. In this case it should return the command so that the process executor can run it.

test/core/test_shell.py
```diff
@@ -0,0 +1,58 @@
...
+    assert capsys.readouterr().out == "[CMD]: ros2 bag play my_bag\n"
+
+
+def test_multiple_args(shell, capsys):
+    """Test that multiple positional arguments are appended in order."""
+    shell.ros2.bag.play("my_bag", "--clock", "-r", "1.0")
+    assert capsys.readouterr().out == "[CMD]: ros2 bag play my_bag --clock -r 1.0\n"
+
+
+def test_kwarg_becomes_flag(shell, capsys):
```
xaru8145, line 46: The kwarg test only covers the `--flag` value case. Missing cases are:

- Boolean True kwarg producing a standalone flag
- Boolean False kwarg being ignored
- A path with spaces as a positional arg e.g. `shell.ros2.bag.play("/my path/to/bag.mcap")`
- Underscore to hyphen conversion with multiple underscores e.g. `save_all_results`

src/lambkin/core/shell/proxy.py
```diff
@@ -17,3 +17,50 @@
...
+
+        for arg in args:
+            extra.append(str(arg))
+
+        for key, value in kwargs.items():
+            flag = "--" + key.replace("_", "-")
+            if value is True:
+                extra.append(flag)
+            elif value is not False:
+                extra.extend([flag, shlex.quote(str(value))])
```
xaru8145, line 60: `shlex.quote` is only applied to kwarg values but not to positional args. This could produce incorrect commands if paths contain spaces, e.g. `shell.ros2.bag.play("/my path/bag") `would print `[CMD]: ros2 bag play /my path/bag.`

src/lambkin/core/shell/proxy.py
```diff
@@ -17,3 +17,50 @@
...
+        """Finalize and print the command.
+
+        Positional args are appended as plain tokens. Keyword args are
+        converted to --flag value pairs, with underscores replaced by hyphens.
+        Boolean True values produce a standalone flag, False values are ignored.
+        """
+        extra = []
+
+        for arg in args:
+            extra.append(str(arg))
```
xaru8145:
```suggestion
            extra.append(shlex.quote(str(arg)))
```

xaru8145 (changes requested):

src/lambkin/core/shell/proxy.py
teresa-ortega: Solved in 1022aef

src/lambkin/core/shell/proxy.py
teresa-ortega: Solved in 1022aef

src/lambkin/core/shell/proxy.py
teresa-ortega: Regarding the dry_run flag, I see a few options for how to expose it:
* Option 1 — Environment variable (change in context file) Set it via an env var in thecontext where ctx.shell is instantiated:
```python
pythonctx.shell = ShellProxy(dry_run=os.getenv("DRY_RUN", "true").lower() == "true")
```
Files to change: **context.py, shell/proxy.py**
* Option 2 — CLI option (change in benchmark script) Expose it as a lambkin option in the benchmark script:
```python
python@lambkin.option("--dry-run", default=True)
def nominal(ctx):
    ctx.shell = ShellProxy(dry_run=ctx.options.dry_run)
```
Files to change: **examples/beluga/beluga_benchmark.py, shell/proxy.py**
Which approach do you prefer?

src/lambkin/core/shell/proxy.py
teresa-ortega: Solved in cd71b24

src/lambkin/core/shell/proxy.py
xaru8145: As agreed via chat, let's go with option 2

teresa-ortega: I choose the option 2, implemented in 07be477

test/core/test_shell.py
teresa-ortega: Solved in 1ceb415

examples/beluga/beluga_benchmark.py
```diff
@@ -24,6 +24,7 @@
 )
 @lambkin.option("--clock-rate", default=100.0)
 @lambkin.option("--sensor-topic", default="/scan")
+@lambkin.option("--dry-run", default=False)
```
xaru8145: Wouldn't we want the default to be True as it is the only supported option right now?

src/lambkin/core/shell/proxy.py
```diff
@@ -17,3 +17,62 @@
...
+            if value is True:
+                extra.append(flag)
+            elif value is not False:
+                extra.extend([flag, shlex.quote(str(value))])
+
+        command = " ".join(self._parts + extra)
+
+        if self._dry_run:
+            # TODO: extend to dispatch commands to BackgroundProcess for real execution.
+            print(f"[CMD]: {command}")
```
xaru8145, line 66: What happens if it is not a dry run? Right now it silently does nothing. Add a print or throw stating that non-dry run is currently not supported.

xaru8145 (changes requested):

examples/beluga/beluga_benchmark.py
teresa-ortega: Yes, you are right, my bad! I was still doing some tests and forgot to change it back. Fixed in 303b756

src/lambkin/core/shell/proxy.py
teresa-ortega: Solved in 303b756

examples/beluga/beluga_benchmark.py
```diff
@@ -39,6 +40,7 @@ def nominal(ctx):
     print(f"sensor_topic: {ctx.options.sensor_topic}")
     print(f"source: {ctx.source}")
     print("---")
+    ctx.shell = lambkin.ShellProxy(dry_run=ctx.options.dry_run)
```
xaru8145: Why is this needed? We shouldn't create the shell ourselves in the nominal function since it is created on the context object. Therefore, it should be able to access the options directly and pass the dry run mode option.

src/lambkin/core/ctx/context.py
```diff
@@ -176,7 +178,7 @@ def __init__(
         # as parameters.
 
         # ctx.shell
-        # TODO(teresa-ortega): self.shell = Shell(self)
+        self.shell = ShellProxy()
```
```suggestion
        self.shell = ShellProxy(dry_run=self.options.dry_run)
```

xaru8145 (changes requested):
Almost there, final fix required.

src/lambkin/core/ctx/context.py
teresa-ortega: I implemented it a bit differently. If the user doesn't define it, the previous approach raises an error because ctx.options.dry_run doesn't exist (since it's just a simple namespace). With this implementation:
```python
ctx.shell = ShellProxy(dry_run=getattr(ctx.options, 'dry_run', False))
``` 
it runs perfectly.
Implemented in d949e0b

examples/beluga/beluga_benchmark.py
teresa-ortega: Solved in d949e0b

src/lambkin/core/ctx/context.py
xaru8145: Okay, fine for now. We'll rework this using SDK flags in next phase and this will not be needed.

xaru8145 (approved):
LGTM.

Now that we have worked on the foundation of the SDK, it's a good idea to start creating tests that combine the different elements to assert hidden issues and verify integration. 

We want to avoid having to manually test everything via the beluga benchmark example.

PR #148, opened 2026-04-29
Phase 4: create metadata files

teresa-ortega:
### Proposed changes

Add YAML metadata file generation to each benchmark iteration output folder.
At context creation, _write_metadata() writes metadata.yaml with run identity, variant parameters, options, source path, output paths, and start timestamp.

#### Type of change

💥 No breaking changes

### Checklist

### Additional comments

N/A

### DIFF
--- a/src/lambkin/core/decorators/benchmark.py
+++ b/src/lambkin/core/decorators/benchmark.py
@@ -84,6 +84,12 @@ def benchmark(variants, num_iterations):
             cli_args = sys.argv[1:] if args is None else args
             options = _parse_options(fn, cli_args)
             source = Source(path=inspect.getfile(fn))
+            # The base context creates a directory for variant 1 / iteration 1
+            # containing a metadata file with the information available at this
+            # point in time.
+            # This directory will later be overwritten with the
+            # actual data collected for variant 1 / iteration 1 during
+            # execution.
             base_ctx = Context(
                 variant={},
                 iteration=0,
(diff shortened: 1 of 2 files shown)


xaru8145 (commented):
Looks good! Several aspects to consider for the future:
- Add a state in the metadata file (started/in progress/finished). But that will depend on how we decide to implement partial restarts and caching. So fine to leave out for now.
- Consider other alternatives like a per benchamrk or per variant metadata to reduce the amount of data dumped. Usage and experience will tell use exactly what we need so we can proceed as is.

One thing to consider now @teresa-ortega: What happens with the base context we use for fetching the inputs? It has invalid variant/iterations, so what will happen with its metadata? Commenting rather than approving as this aspect is relevant to sort out first,

teresa-ortega:
I propose using `Context.__new__(Context)` for the `base_ctx`, as it does not call `__init__` and therefore no folders are created on disk nor invalid metadata is generated.

```python
base_ctx = Context.__new__(Context)
```

Note: for this implementation to work correctly, the contents of PR #149 are required.

xaru8145:
> I propose using `Context.__new__(Context)` for the `base_ctx`, as it does not call `__init__` and therefore no folders are created on disk nor invalid metadata is generated.
> 
> ```python
> base_ctx = Context.__new__(Context)
> ```
> 
> Note: for this implementation to work correctly, the contents of PR #149 are required.

This seems a bit hacky to me. I would prefer passing a new flag to the constructor like `write_metadata` for now.

src/lambkin/core/decorators/benchmark.py
```diff
@@ -84,6 +84,12 @@ def wrapper(args=None, output_dir=None):
             cli_args = sys.argv[1:] if args is None else args
             options = _parse_options(fn, cli_args)
             source = Source(path=inspect.getfile(fn))
+            # The base context creates a directory for variant 1 / iteration 1
+            # containing a metadata file with the information available at this
```
xaru8145, line 88: Be explicit about this information being dummy
```suggestion
            # containing a metadata file with dummy information available at this
```

xaru8145 (approved):
LGTM minor comment.

PR #149, opened 2026-04-30
Phase 4: Make Context immutable

teresa-ortega:
### Proposed changes

Previously, any part of the code could mutate ctx freely, making the execution flow hard to reason about. This PR decorates Context with @dataclass(frozen=True) so any field reassignment raises a FrozenInstanceError. Since the dataclass blocks writes in __init__ too, internal assignments use object.__setattr__, which is the standard Python pattern for frozen dataclasses with custom construction logic. A new test test_context_is_immutable verifies the behaviour.

#### Type of change

💥 No breaking changes

### Checklist

### Additional comments
N/A

### DIFF
--- a/src/lambkin/core/ctx/context.py
+++ b/src/lambkin/core/ctx/context.py
@@ -81,26 +81,29 @@ def _iteration_folder_name(iteration: int) -> str:
 class Context:
     """Carries all namespaced information for one benchmark variant.
 
-    Builds all namespaced sub-objects (variant, inputs, options, output)
-    from the given parameters and automatically creates the required output
-    folders on disk.
+    Immutable by design: all fields are set once at construction time
+    and cannot be reassigned afterwards. Any attempt to set or delete
+    an attribute after construction raises an AttributeError.
 
     Attributes:
     ----------
-    BENCHMARKS_DIRNAME : str
-        Name for the benchmarks directory.
-    variation:
+    variant : SimpleNamespace
         Namespaced algorithm parameters for this run.
         All key-value pairs from the variant dict are exposed as attributes.
-    source:
-        Source object describing the benchmark script being executed.
-    inputs:
-        Namespaced input information.
-    options:
+    options : SimpleNamespace
         Namespaced runtime options.
         All key-value pairs from the options dict are exposed as attributes.
-    output:
-        Namespaced output paths.
+    inputs : SimpleNamespace
+        Namespaced input information, injected by the benchmark runner
+        after resolving all registered input hooks.
+    iteration : int
+        Zero-based repetition index within this variant.
+    source : Source
+        Source object describing the benchmark script being executed.
+    output : OutputInfo
+        Namespaced output paths for this variant and iteration.
+    shell : ShellProxy
+        Shell proxy for running commands during the benchmark.
     """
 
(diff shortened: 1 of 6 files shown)


src/lambkin/core/decorators/benchmark.py
```diff
@@ -104,7 +104,7 @@ def wrapper(args=None, output_dir=None):
                         variant_index=variant_index,
                         output_dir=output_dir,
                     )
-                    ctx.inputs = resolved_inputs
+                    object.__setattr__(ctx, "inputs", resolved_inputs)
```
xaru8145: This inmutability fix trades one problem for another: `frozen=True` advertises immutability
but `object.__setattr__` bypasses it just as easily as a plain assignment. 

Consider accepting `inputs` as an optional constructor.

FYI https://python-course.eu/oop/creating-immutable-classes-in-python.php

xaru8145 (changes requested):

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in eb061b4

xaru8145 (changes requested):
This solution trades one bypass for another. Instead of `object.__setattr__(self, ...)`, you have `obj.__dict__.update(...)`. Both are equally accessible to user code.

So it doesn't improve the actual immutability guarantee over the `__setattr__` override approach. It just moves the initialization pattern from object.`__setattr__` to `__dict__` manipulation. There's also a subtle downside: moving all construction logic into `__new__` is unusual and slightly surprising to readers .

My previous review referenced the use of setattr in the benchmark, not the context.

The core idea here is to avoid users from doing `ctx.variant = my_variant' on nominal. If the user tried to hack by doing __setattr__ on the nominal function, we can't do anything about it. When we say inmutable we are really just preventing accidental mutation.

src/lambkin/core/ctx/context.py
```diff
@@ -139,25 +142,19 @@ def __init__(
...
+        object.__setattr__(self, "inputs", inputs)
 
         # ctx.iteration
-        self.iteration = iteration
-
-        # TODO: Consider moving path construction logic into OutputInfo itself,
-        # giving it a constructor that takes base_dir, variation_index, and
-        # iteration and derives variation_dir and iteration_dir internally. This
-        # would make OutputInfo a cohesive object that owns everything
-        # path-related.
```
xaru8145, line 163: Why is this removed? Isn't it still a valid TODO?

xaru8145 (changes requested):
Minor things. Almost done.

Wouldn't it be better to update `test_input.py`so that it uses the returned object of resolve function to make the assertions, for example here:
https://github.com/Ekumen-OS/lambkin/blob/369c5feeef2496f95a44079b2667dd95dfdc5cde/test/core/test_input.py#L49-L50
Rather than accessing a hidden element of the inputs object.

src/lambkin/core/ctx/context.py
teresa-ortega: Yes it is still a valid TODO,it was an accidental error. Solved in d8bcba3

xaru8145 (approved):

PR #151, opened 2026-05-13
Phase 4: Background Process Lifecycle

teresa-ortega:
### Proposed changes

This PR introduces a secure context manager for running background processes using cgroups v2.
By isolating each command in a dedicated child cgroup, we guarantee that no orphan processes are left behind.
It implements a graceful shutdown sequence, sending SIGTERM followed by SIGKILL after a grace period.
Additionally, a dedicated background thread monitors the process lifecycle in real-time.
If the process dies unexpectedly before the context manager exits, it raises a custom error to fail fast and aid debugging.

#### How to test 
The following manual tests exercise the full benchmark stack including background process management via cgroups v2. Two container runtimes are supported.

##### Prerequisites
Mount your rosbag, map, and ground truth by adding the corresponding volumes to the run command:
```
-v /path/to/rosbag:/ws/examples/beluga/rosbags:ro
-v /path/to/map:/ws/examples/beluga/maps:ro
-v /path/to/groundtruth:/ws/examples/beluga/groundtruth:ro
```

Build the image:
```bash
cd examples/beluga/docker
docker compose build lambkin_prod  # Docker
podman compose build lambkin_prod  # Podman (alternatively)
```

##### Podman
```bash
podman run --rm \
  --name=lambkin_production \
  --systemd=always \
  --network=host \
  -v "$(pwd)/../results:/ws/examples/beluga/results" \
  -it lambkin_ros:jazzy bash
```
Inside the container:
```bash
```
##### Docker
```bash
docker compose -f docker-compose.yml --profile production run --rm lambkin_prod bash
```

Inside the container:
```bash
```
#### Integration tests
The following scripts test the background process lifecycle in isolation without requiring the full ROS 2 stack. Run them with:
```bash
uv run lambkin <script>
```
**Test 1 — Normal execution:**
```bash
uv run lambkin  test/integration/integration_test_normal_execution.py
```
Verifies that two nested background processes start correctly, the foreground runs to completion, and both background processes terminate in reverse order with their cgroups cleaned up.

**Test 2 — Unexpected death of a background process:**
```bash
uv run lambkin  test/integration/integration_test_unexpected_death.py
```
Verifies that when an inner background process dies before the context manager exits, LambkinProcessDiedUnexpectedlyError is raised and the outer background process is also terminated.
**Test 3 — Foreground failure:**
```bash
uv run lambkin test/integration/integration_test_foreground_fails.py
```
Verifies that when the foreground raises an exception, both background processes are terminated cleanly and the original exception is propagated.
**Test 4 — Intentional exception in a background process:**
```bash
uv run lambkin test/integration/integration_test_exception.py
```
Verifies that when a background process raises an unhandled exception and exits with a non-zero return code, LambkinProcessDiedUnexpectedlyError is raised and all other background processes are terminated.
**Test 5 — Cgroup tree cleanup after iteration:**
```bash
uv run lambkin test/integration/integration_test_cgroups_cleanup.py
```
Verifies that after a `Context` exits normally, the process cgroup and iteration cgroup are fully removed from the filesystem.
#### Manual validation
**Test 1 — Kill a background process and verify the whole tree is cleaned up.**
In one terminal launch the benchmark and in another terminal kill one of the background processes:
```bash
ps aux | grep -E "ros2|record|launch" | grep -v grep
kill -9 <PID>
ps aux | grep -E "ros2|record|launch" | grep -v grep  # should return empty
```
**Test 2 — Launch all three processes and verify they wait and terminate cleanly.**
While the benchmark runs, verify in another terminal that the three processes are running simultaneously:
```bash
watch -n 1 'ros2 node list'
```
When play finishes, verify that launch stops first and then record stops in that order, and that no processes remain.

#### Type of change

💥 No breaking changes

### Checklist

### Additional comments

N/A


src/lambkin/core/process/__init__.py
```diff
@@ -18,3 +18,7 @@
 - background: non-blocking process launch.
 - foreground: blocking process execution.
```
xaru8145: I think we can remove the `foreground` comment.

src/lambkin/core/ctx/context.py
```diff
@@ -189,13 +190,20 @@ def __init__(
 
         self._setup_directories()
 
+        iteration_cgroup = make_iteration_cgroup(
+            find_delegated_cgroup(),
+            iteration_dir,
+        )
+        object.__setattr__(self, "iteration_cgroup", iteration_cgroup)
```
xaru8145: Why do we need to make it an attribute of the class? Where is it used elsewhere?

src/lambkin/core/process/background.py
```diff
@@ -18,3 +18,203 @@
...
+            The command that died.
+        returncode : int
+            The exit code of the process.
+        """
+        self.argv = argv
+        self.returncode = returncode
+        super().__init__(
+            f"Background process {argv[0]!r} died unexpectedly "
+            f"with return code {returncode}."
+        )
```
xaru8145: Do you think it could make sense to move `ProcessDiedUnexpectedlyError` to `exceptions.py`?

src/lambkin/core/process/background.py
```diff
@@ -18,3 +18,203 @@
...
+
+import os
+import subprocess
+import threading
+import uuid
+from pathlib import Path
+from typing import Any
+
+from lambkin.core.process.cgroup import kill_cgroup, make_cgroup, remove_cgroup
+from lambkin.core.shell.proxy import _CommandProxy
```
xaru8145: `background()` is a public API function but its signature relies on `_CommandProxy`, a private class from another module. This leaks an implementation detail into the public interface. 

Consider promoting to `CommandProxy` as it seems it is not private any more but intended to be used by other classes like background.

Let's try to keep in mind all the naming conventions and everything for project consistency.

src/lambkin/core/shell/proxy.py
xaru8145: Would drop the leading underscore in `_CommandProxy` as it is used in background so it is no longer sth privetly operated by Shell. It is actually what is returned when you construct a shell command.

Same as mentioned here https://github.com/Ekumen-OS/lambkin/pull/151#discussion_r3282484063

src/lambkin/core/process/__init__.py
xaru8145, line 1: cgroup utilities not exported, inconsistent public surface

src/lambkin/core/process/background.py
```diff
@@ -18,3 +18,203 @@
...
+
+GRACE_PERIOD = 3.0
```
xaru8145: Move `GRACE_PERIOD` from `background.py` to `common/defaults.py`. Suggest renaming it `SIGTERM_GRACE_PERIOD` to make clear what it applies to.

src/lambkin/core/process/cgroup.py
```diff
@@ -0,0 +1,141 @@
...
+            return
+        time.sleep(0.05)
+
+    for pid_str in procs_file.read_text().split():
+        try:
+            os.kill(int(pid_str), signal.SIGKILL)
+        except ProcessLookupError:
+            pass
+
+    deadline = time.monotonic() + 5.0
```
xaru8145: `kill_cgroup` and `remove_cgroup` have a hardcoded `5.0` timeout that should be extracted to `common/defaults.py` as `SIGKILL_GRACE_PERIOD`. As-is it's a magic number with no explanation.

src/lambkin/core/process/cgroup.py
```diff
@@ -0,0 +1,141 @@
...
+    while time.monotonic() < deadline:
+        if not procs_file.read_text().strip():
+            return
+        time.sleep(0.05)
```
xaru8145: Same here. Maybe define as `CGROUP_POLL_INTERVAL` constant?

Let's try to avoid hardcoded magical numbers.

src/lambkin/core/process/background.py
```diff
@@ -18,3 +18,203 @@
...
+
+    def _make_cgroup(self) -> Path:
+        """Create a unique child cgroup under the iteration cgroup.
+
+        Returns:
+        -------
+        Path
+            The path to the newly created cgroup directory.
+        """
+        name = f"{self._argv[0].split('/')[-1]}-{uuid.uuid4().hex[:8]}"
```
xaru8145: There are two separate places creating cgroup names with different conventions and different amounts of information embedded in the name. `BackgroundProcess._make_cgroup` creates a child cgroup named after the process binary, while `make_iteration_cgroup` creates one named after the iteration directory structure. Both use a UUID suffix for uniqueness.

The inconsistency raises a question: should cgroup naming live in one place? `BackgroundProcess._make_cgroup` is a private method that duplicates the responsibility of `make_cgroup` + a naming convention that arguably belongs in `cgroup.py`. It could simply be replaced by a call to a dedicated `make_process_cgroup(parent, argv)` function in `cgroup.py`, keeping all cgroup naming logic in one module.

xaru8145 (changes requested):
First pass.

src/lambkin/core/process/__init__.py
teresa-ortega: Solved in f9d82f8

src/lambkin/core/ctx/context.py
teresa-ortega: You're right, it's unnecessary. Solved in b5e8c8f

src/lambkin/core/process/background.py
teresa-ortega: Solved in 2a3bc59

src/lambkin/core/process/__init__.py
teresa-ortega: After giving it some thought, `cgroup.py` is purely an implementation detail — its functions are only ever called from `background.py` and the user has no reason to access them directly. Given that, I think the right fix is to prefix them with an underscore (`_kill_cgroup`, `_make_cgroup`, `_remove_cgroup`) to make their private nature explicit, and leave them out of any `__init__.py`. This way the public surface stays clean and consistent.

src/lambkin/core/process/background.py
teresa-ortega: Solved in 1ff8be4

src/lambkin/core/shell/proxy.py
teresa-ortega: SOlved in 1ff8be4

src/lambkin/core/process/cgroup.py
teresa-ortega: Solved in 2ceb354

src/lambkin/core/process/cgroup.py
teresa-ortega: Solved in 2ceb354

src/lambkin/core/process/background.py
teresa-ortega: Solved in 6247dad

src/lambkin/core/process/background.py
teresa-ortega: Solved in 4cd0420

src/lambkin/core/process/__init__.py
teresa-ortega: Solved in d5d507a

src/lambkin/core/process/__init__.py
```diff
@@ -14,7 +14,30 @@
...
+    kill_cgroup,
+    make_cgroup,
+    make_iteration_cgroup,
+    make_process_cgroup,
+    remove_cgroup,
+)
+
+__all__ = [
+    "BackgroundProcess",
+    "LambkinProcessDiedUnexpectedlyError",
```
xaru8145: Why is this exported here? Should be done only in common __init__ file.

test/core/test_background.py
```diff
@@ -0,0 +1,201 @@
...
+def test_background_passes_dry_run_from_proxy(fake_cgroup, tmp_path):
+    """background() passes the dry_run flag from the proxy."""
+    shell = ShellProxy(dry_run=True, cwd=tmp_path, cgroup=fake_cgroup)
+    bp = background(shell.sleep, "10")
+    assert bp._dry_run is True
+
+
+def test_background_process_starts_and_stops(tmp_path):
+    """BackgroundProcess starts a real process and kills it on exit."""
+    from lambkin.core.process.cgroup import find_delegated_cgroup, make_iteration_cgroup
```
xaru8145: All the imports should happen at the top of the file.

test/core/test_cgroup.py
```diff
@@ -0,0 +1,210 @@
...
+    iteration_dir = tmp_path / "var_1" / "iter_1"
+    iteration_dir.mkdir(parents=True)
+    result1 = make_iteration_cgroup(tmp_path, iteration_dir)
+    result2 = make_iteration_cgroup(tmp_path, iteration_dir)
+    assert result1 != result2
+
+
+def test_kill_cgroup_terminates_cooperative_process():
+    """kill_cgroup sends SIGTERM and the process exits cleanly."""
+    from lambkin.core.process.cgroup import find_delegated_cgroup
```
xaru8145: Imports should happen in the top of the file.

src/lambkin/core/process/cgroup.py
```diff
@@ -0,0 +1,162 @@
...
+    ------
+    RuntimeError
+        If no cgroup v2 directory is found.
+    """
+    cgroup_file = Path("/proc/self/cgroup")
+    for line in cgroup_file.read_text().splitlines():
+        if line.startswith("0::"):
+            rel = line[3:].strip()
+            return Path("/sys/fs/cgroup") / rel.lstrip("/")
+    raise RuntimeError("No cgroup v2 found.")
```
xaru8145: `find_delegated_cgroup` returns the path without verifying the process has write access to it. If the container isn't set up with proper cgroup delegation, the failure will surface in an unexpected place with a confusing error rather than here.
Commenting for awareness. If the message thrown is descriptive enough, fine as is. Otherwise, consider adding a check and throw intentionally with a meaningful and relevant message.

src/lambkin/core/process/cgroup.py
```diff
@@ -0,0 +1,162 @@
...
+        try:
+            os.kill(int(pid_str), signal.SIGKILL)
+        except ProcessLookupError:
+            pass
+
+    deadline = time.monotonic() + defaults.SIGKILL_GRACE_PERIOD
+    while time.monotonic() < deadline:
+        if not procs_file.read_text().strip():
+            return
+        time.sleep(defaults.CGROUP_POLL_INTERVAL)
```
xaru8145, line 144: After the SIGKILL grace period, `kill_cgroup` returns silently even if processes are still alive in the cgroup.  Several questions/comments:
- What happens if we attempt to `remove_cgroup` afterwards in this case? Could it fail silentlty?
- A process surviving SIGKILL indicates a serious system problem so silently returning could be risky. We should consider what to do. Raising here feels to defensive, so at least let's print to terminal.

src/lambkin/core/process/cgroup.py
```diff
@@ -0,0 +1,162 @@
...
+    cgroup : Path
+        The cgroup directory to remove.
+    """
+    deadline = time.monotonic() + defaults.SIGKILL_GRACE_PERIOD
+    while time.monotonic() < deadline:
+        try:
+            cgroup.rmdir()
+            return
+        except OSError:
+            time.sleep(defaults.CGROUP_POLL_INTERVAL)
```
xaru8145: Same comment as `kill_cgroup`.

src/lambkin/core/process/cgroup.py
```diff
@@ -0,0 +1,162 @@
...
+    ----------
+    cgroup : Path
+        The cgroup directory to remove.
+    """
+    deadline = time.monotonic() + defaults.SIGKILL_GRACE_PERIOD
+    while time.monotonic() < deadline:
+        try:
+            cgroup.rmdir()
+            return
+        except OSError:
```
xaru8145: This error might be a bit broad. Are we sure we don't want to be more specific?

src/lambkin/core/process/background.py
```diff
@@ -18,3 +18,162 @@
...
+
+        Runs in the child process after fork() but before exec().
+        """
+        (self._cgroup / "cgroup.procs").write_text(str(os.getpid()))
+
+    def _monitor_process(self) -> None:
+        """Monitor thread that detects if the process dies unexpectedly."""
+        self._proc.wait()
+        if not self._exiting:
+            self._died_unexpectedly = True
```
xaru8145, line 87: `self._exiting` is set in `__exit__` (main thread) and read concurrently in `_monitor_process` (monitor thread). When accessing a variable from different threads there can be race conditions.

In CPython the [GIL](https://wiki.python.org/moin/GlobalInterpreterLock) prevents accesing objects at the same time from different threads, but it's not idiomatic and not guaranteed by the language spec. You can research a bit more about this. 

According to the [Google Style guide on Python](https://github.com/google/styleguide/blob/91d6e367e384b0d8aaaf7ce95029514fcdf38651/pyguide.md#218-threading), it recommends using `Queue` or `threading` module instead. I would probably go for the second one.

src/lambkin/core/process/background.py
```diff
@@ -18,3 +18,162 @@
...
+        kill_cgroup(self._cgroup, grace_period=defaults.SIGTERM_GRACE_PERIOD)
+
+        if self._monitor is not None:
+            self._monitor.join()
+
+        remove_cgroup(self._cgroup)
+
+        if exc_type is None and self._died_unexpectedly:
+            returncode = self._proc.poll()
+            raise LambkinProcessDiedUnexpectedlyError(self._argv, returncode or 1)
```
xaru8145: What happens if exc_type is not none of the process we are running in background? Will we still kill and remove the cgroups?

xaru8145 (changes requested):
I am missing the use of `lambkin.process.background` in beluga example benchmark.

test/core/test_background.py
teresa-ortega: Solved in a8ff256

test/core/test_cgroup.py
teresa-ortega: Solved in a8ff256

src/lambkin/core/process/cgroup.py
teresa-ortega: Solved in 5cb61ec

src/lambkin/core/process/cgroup.py
teresa-ortega: Good points. To answer both questions:
* Regarding remove_cgroup , yes, it would fail silently in the current implementation. It retries for 5 seconds and then returns without any indication that the cgroup was not removed, which means leaked processes and leaked cgroup directories would go unnoticed.
* Regarding processes surviving SIGKILL , agreed that raising feels too defensive since this is a kernel-level problem that the benchmark cannot recover from. A warning to stderr is the right approach: it makes the problem visible without stopping execution abruptly in a situation where the system itself is misbehaving.
The proposed fix is to add a warning to stderr in both cases ,in kill_cgroup if processes survive SIGKILL, and in remove_cgroup if the directory cannot be removed after the timeout. This way the user is always informed when something goes wrong at the cgroup level, without the benchmark failing on what is ultimately a system problem.

teresa-ortega: Solved in 98ee65d

src/lambkin/core/process/cgroup.py
teresa-ortega: solved in 98ee65d

src/lambkin/core/process/cgroup.py
teresa-ortega: We can be more specific by checking errno , retrying only on ENOTEMPTY, treating ENOENT as a silent success, and propagating any other OSError.

teresa-ortega: Solved in aeea0cc

src/lambkin/core/process/background.py
teresa-ortega: Replacing self._exiting with a threading.Event is a very good approach. Solved in f7c4059

src/lambkin/core/process/background.py
teresa-ortega: Yes, kill_cgroup and remove_cgroup are always executed regardless of exc_type, the cleanup happens unconditionally. The if exc_type is None check only controls whether LambkinProcessDiedUnexpectedlyError is raised, to avoid masking an already propagating exception.

src/lambkin/core/process/__init__.py
teresa-ortega: Solved in f1f51d0

examples/beluga/beluga_benchmark.py
```diff
@@ -33,14 +33,22 @@ def nominal(ctx):
...
-    ctx.shell.evo_ape.bag2("output.mcap", save_results="out.zip")
+    with lambkin.process.background(
+        ctx.shell.ros2.bag.record,
+        "-O",
+        "output.mcap",
+        "-a",
+    ):
+        with lambkin.process.background(
+            ctx.shell.ros2.launch,
+            ctx.source.path.parent / "beluga.launch.xml",
```
xaru8145: Shouldn't this be a `.launch.py` extension rather than `.launch.xml`?

```suggestion
            ctx.source.path.parent / "beluga.launch.py",
```

Have you actually tested this with real rosbags?

teresa-ortega: Solved in f282fb8

examples/beluga/beluga_benchmark.py
xaru8145: Let's use a rate of 1 here @teresa-ortega

teresa-ortega: Solved in cec839f

src/lambkin/core/process/cgroup.py
```diff
@@ -0,0 +1,198 @@
...
+            candidate = Path("/sys/fs/cgroup") / rel.lstrip("/")
+            if not candidate.exists():
+                raise RuntimeError(
+                    f"cgroup v2 directory not found: {candidate}. "
+                    "Make sure cgroup v2 is enabled on this system."
+                )
+            if not os.access(candidate, os.W_OK):
+                raise RuntimeError(
+                    f"No write access to cgroup: {candidate}. "
+                    "Run with: systemd-run --user --scope, "
```
xaru8145: I do not think we should encourage the user to run `systemd-run --user --scope` as this is handled internally via `uv run lambkin`.

src/lambkin/core/process/cgroup.py
```diff
@@ -0,0 +1,198 @@
...
+    cgroup : Path
+        The cgroup directory to remove.
+    """
+    deadline = time.monotonic() + defaults.SIGKILL_GRACE_PERIOD
+    while time.monotonic() < deadline:
+        try:
+            cgroup.rmdir()
+            return
+        except OSError as e:
+            if e.errno == errno.ENOTEMPTY:
```
xaru8145: Are you sure that you will get this? Is it possible that we can get `EBUSY`?

test/core/test_background.py
```diff
@@ -0,0 +1,182 @@
...
+    """Return a ShellProxy in dry-run mode with a fake cgroup."""
+    return ShellProxy(dry_run=True, cwd=tmp_path, cgroup=fake_cgroup)
+
+
+def test_background_dry_run_prints_command(dry_shell, capsys):
+    """In dry-run mode, background prints the command instead of executing it."""
+    with background(dry_shell.sleep, "10"):
+        pass
+    out = capsys.readouterr().out
+    assert "[DRY RUN BG]" in out
```
xaru8145, line 48: For maintainability consider wrapping this into a constant. Can be done in a follow up PR or in the logging one.

src/lambkin/core/process/background.py
```diff
@@ -18,3 +18,164 @@
...
+    with background(ctx.shell.ros2.bag.record, "-O", "output.mcap", "-a"):
+        ...
+    """
+    argv = proxy._build_argv(*args, **kwargs)
+    return BackgroundProcess(
+        argv=argv,
+        iteration_cgroup=proxy._cgroup,
+        cwd=proxy._cwd,
+        dry_run=proxy._dry_run,
+    )
```
xaru8145: `CommandProxy` is now public but `_build_argv`, `_cgroup`, `_cwd`, `_dry_run` are still private. Consider making them public or defining getter function.s

test/integration/integration_test_unexpected_death.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+    iteration_dir = Path("/tmp/integration_test/var_1/iter_1")
+    iteration_dir.mkdir(parents=True, exist_ok=True)
+    cgroup = make_iteration_cgroup(find_delegated_cgroup(), iteration_dir)
+
+    shell = ShellProxy(dry_run=False, cwd=iteration_dir, cgroup=cgroup)
+
+    print("Starting outer background process...")
+    try:
+        with background(shell.__getattr__(sys.executable), "-c", COOPERATIVE) as bp1:
```
xaru8145: Why not use it like this? Which reads more natural and follows typical usage.
```suggestion
        with background(shell.python3, "-c", COOPERATIVE) as bp1:
```

test/integration/integration_test_exception.py
xaru8145, line 1: Do you think all integration tests could benefit from asserting that the cgroup that we create is effectivelly killed and removed? If so, would it make sense to add a function in `cgroup.py` to check if a given cgroup exists?

test/integration/integration_test_exception.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def main():
+    """Run integration test for an intentional exception in a background process.
+
+    Verifies that when a background process raises an unhandled exception and
+    exits with a non-zero return code, LambkinProcessDiedUnexpectedlyError is
+    raised and all other background processes are terminated.
+    """
+    iteration_dir = Path("/tmp/integration_test/var_1/iter_1")
```
xaru8145: All four tests share the same hardcoded path. If two tests run concurrently, or if a previous run left state behind, they'll interfere with each other. Use `tempfile.mkdtemp()` and clean up after, or at least use unique subdirectories per test.

test/integration/integration_test_exception.py
```diff
@@ -0,0 +1,74 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Integration test: background process raises an exception intentionally."""
+
+import sys
+import time
+from pathlib import Path
+
+sys.path.insert(0, "src")
```
xaru8145: Why is this needed? In case of automated tests (near future), pyproject or pytest should handle it. As of now, if LAMBKIN is properly installed via pip install this shouldn't be necessary.

xaru8145 (changes requested):

examples/beluga/beluga_benchmark.py
```diff
@@ -33,26 +33,35 @@ def nominal(ctx):
...
     """Return the path to the MCAP dataset used as input for the benchmark."""
-    return "/mydataset/dataset.mcap"
+    return ctx.source.path.parent / "my_bags" / "my_bag.mcap"
 
 
 @nominal.input
 def map(ctx):
     """Return the path to the map file used for localization."""
-    return "my_map.yaml"
+    return "maps/map.yaml"
```
xaru8145: This does not match the volumes detailed in the PR summary:

```
-v /path/to/rosbag:/ws/examples/beluga/rosbags:ro
-v /path/to/map:/ws/examples/beluga/maps:ro
-v /path/to/groundtruth:/ws/examples/beluga/groundtruth:ro
```

src/lambkin/core/process/cgroup.py
teresa-ortega: Solved in 00d696a

src/lambkin/core/process/cgroup.py
teresa-ortega: You're right. Solved in e552b63

test/core/test_background.py
teresa-ortega: Okay, I will do in the logging PR #156

src/lambkin/core/process/background.py
teresa-ortega: Solved in f8d2588

test/integration/integration_test_foreground_fails.py
xaru8145, line 1: The file name has a typo. "foregound" should be "foreground". Missing an "r".

test/integration/integration_test_unexpected_death.py
teresa-ortega: Solved in 7878969

test/integration/integration_test_exception.py
teresa-ortega: Solved in 30fedb9

test/integration/integration_test_exception.py
teresa-ortega: Solved in 2c5ac27

test/integration/integration_test_exception.py
teresa-ortega: Solved in b055244

examples/beluga/beluga_benchmark.py
teresa-ortega: Solved in 3768e64

test/integration/integration_test_foreground_fails.py
teresa-ortega: Solved in b2dd6c0

examples/beluga/docker/Dockerfile
```diff
@@ -1,9 +1,10 @@
 # ── Stage 0: base ─────────────────────────────────────────
 FROM ros:jazzy-ros-base AS base
-
 RUN apt-get update && apt-get install -y \
     curl \
     python3-pip \
+    systemd \
+    dbus-user-session \
```
xaru8145: I don¡t think we need `dbus-user-session` as it is a dependency of `systemd-run --user`. It's needed to communicate with the user systemd session over D-Bus.

src/lambkin/cli.py
xaru8145, line 1: Why not drop the systemd run also when running in non containerized environments?

On a bare host, `find_delegated_cgroup()` will find the user session scope that systemd creates automatically for every logged-in user, which is already delegated and writable.

The cgroup isolation is fully handled by Context and BackgroundProcess as Context.__init__ calls make_iteration_cgroup(find_delegated_cgroup(), iteration_dir) which creates a dedicated child cgroup for each iteration under the delegated cgroup provided by the environment.

Unless I am missing something, which I am totally open to discuss.

src/lambkin/cli.py
```diff
@@ -86,6 +84,11 @@ def _stop_scope(cgroup_scope: str) -> None:
         ) from err
 
 
+def _running_in_container() -> bool:
+    """Return True if running inside a container."""
+    return Path("/.dockerenv").exists() or Path("/run/.containerenv").exists()
```
xaru8145: This could be more robust (see https://github.com/ysskrishna/container-detect/blob/main/container_detect/core.py) and moved to utils. But depending on how the comment above is resolved we might not need to distinguish if we are inside a container.

xaru8145 (changes requested):
When I run it either in the host or podman, I get:

```bash
[lifecycle_manager-3] [INFO] [1780007182.415736802] [lifecycle_manager_localization]: Server beluga_amcl connected with bond.
[lifecycle_manager-3] [INFO] [1780007182.415769113] [lifecycle_manager_localization]: Managed nodes are active
[lifecycle_manager-3] [INFO] [1780007182.415774550] [lifecycle_manager_localization]: Creating bond timer...
[amcl_node-1] [INFO] [1780007182.416163376] [beluga_amcl]: The bond connection to the lifecycle manager is now fully formed
Traceback (most recent call last):
  File "/ws/examples/beluga/beluga_benchmark.py", line 74, in <module>
    nominal()
  File "/usr/local/lib/python3.12/dist-packages/lambkin/core/decorators/benchmark.py", line 148, in wrapper
    fn(ctx)
  File "/ws/examples/beluga/beluga_benchmark.py", line 36, in nominal
    with lambkin.process.background(
  File "/usr/local/lib/python3.12/dist-packages/lambkin/core/process/background.py", line 148, in __exit__
    raise exceptions.LambkinProcessDiedUnexpectedlyError(
lambkin.common.exceptions.LambkinProcessDiedUnexpectedlyError: Background process 'ros2' died unexpectedly with return code 1.
```

Also, as you can see, the error is not very informative as we can't know if process 'ros2' is the launch, the play or the record. This should be improved.

examples/beluga/docker/Dockerfile
teresa-ortega: Solved in a675321

src/lambkin/cli.py
teresa-ortega: Solved in d028df8

src/lambkin/cli.py
teresa-ortega: This function has been removed entirely ,we no longer need to distinguish between container and host environments

src/lambkin/cli.py
```diff
@@ -65,84 +65,41 @@ def format_help(self, ctx: click.Context, formatter: HelpFormatter) -> None:
...
-                "--user",
-                sys.executable,
-                str(script),
-                *args,
-            ],
-        )
         proc.wait()
     except KeyboardInterrupt:
-        _stop_scope(cgroup_scope)
+        kill_cgroup(find_delegated_cgroup())
```
xaru8145: Isn't this problematic? If I understand it correctly, this will kill the cgroup that runs in the host for the user, so next time you run it there will be no cgroup and `find_delegated_cgroup` could fail or keep joining user cgroups until it kills all of them. This is for the case of running in the host. For the containers, not sure what can happen then.

src/lambkin/cli.py
```diff
@@ -65,84 +65,41 @@ def format_help(self, ctx: click.Context, formatter: HelpFormatter) -> None:
...
 
-    cgroup_scope = f"lambkin-{script.stem}.scope"
-
     def _handle_sigint(signum, frame):
         raise KeyboardInterrupt
 
     signal.signal(signal.SIGINT, _handle_sigint)
+    proc = subprocess.Popen(
+        [sys.executable, str(script), *args],
+        start_new_session=True,
```
xaru8145, line 113: Why do we need to start a new session here?

src/lambkin/cli.py
teresa-ortega: Solved in 9d44086

src/lambkin/cli.py
teresa-ortega: `start_new_session=True` is needed to prevent the Ctrl-C SIGINT from being delivered directly to the benchmark script. Without it, the terminal sends SIGINT to the entire process group, so the script would receive it and crash mid-execution with a `KeyboardInterrupt` traceback. With it, only the CLI receives the SIGINT and kills the child cleanly via `kill_cgroup`.

src/lambkin/cli.py
```diff
@@ -95,11 +104,13 @@ def _handle_sigint(signum, frame):
...
         [sys.executable, str(script), *args],
         start_new_session=True,
+        preexec_fn=lambda: (run_cgroup / "cgroup.procs").write_text(str(os.getpid())),
     )
     try:
         proc.wait()
     except KeyboardInterrupt:
-        kill_cgroup(find_delegated_cgroup())
+        kill_cgroup(run_cgroup)
+        remove_cgroup_tree(run_cgroup)
```
xaru8145: remove_cgroup_tree already calls cgroup.kill internally, so calling kill_cgroup before it is redundant. remove_cgroup_tree handles the full cleanup on its own.

maybe rename or document that remove_cgroup_tree also does the killing. Can be confusing as we have `remove_cgroup` and `kill_cgroup`.

src/lambkin/cli.py
xaru8145: I think docu needs to be updated

src/lambkin/cli.py
```diff
@@ -87,6 +94,8 @@ def main(script: Path, args: tuple) -> None:
     # TODO(teresa-ortega): Handle concurrent runs, interrupted benchmarks, and re-runs
     # (e.g. detect an already active scope, support partial restarts).
     # To be addressed in phase 6.
+    delegated = find_delegated_cgroup()
```
xaru8145: A more intentional approach would be to always create run_cgroup under a well-known, stable parent — something like user@<uid>.service/app.slice which is the conventional place for user applications on a systemd host. That way lambkin cgroups always appear in a predictable location regardless of how the CLI was launched.

But this would require writing to a path the user may not have delegation for, so you'd need to verify that first — or fall back to find_delegated_cgroup() if the preferred parent isn't writable.

src/lambkin/core/decorators/benchmark.py
```diff
@@ -131,21 +132,22 @@ def wrapper(args=None, output_dir=None):
                 variant_index=0,
                 output_dir=output_dir,
             )
+            remove_cgroup(base_ctx._iteration_cgroup)
```
xaru8145: While the loop correctly uses with Context(...) as ctx, base_ctx bypasses the context manager entirely. Inconsistent. Should be:
```py
with Context(...) as base_ctx:
    resolved_inputs = inputs.resolve(base_ctx)
    ...
```

src/lambkin/core/ctx/context.py
```diff
@@ -250,3 +256,11 @@ def __repr__(self) -> str:
...
             f")"
         )
+
+    def __enter__(self) -> Context:
+        """Enter the context manager, returning this instance."""
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+        """Exit the context manager and remove the iteration cgroup."""
+        remove_cgroup(self._iteration_cgroup)
```
xaru8145: Is this safe to do or should we kill the entire tree?

test/integration/integration_test_cgroups_cleanup.py
```diff
@@ -0,0 +1,60 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Integration test: cgroup tree is fully removed after iteration ends."""
+
+import sys
+from pathlib import Path
+
+sys.path.insert(0, "src")
```
xaru8145: This is not needed as we agreed on the other integration tests.

test/integration/integration_test_cgroups_cleanup.py
```diff
@@ -0,0 +1,60 @@
...
+    source = Source(path=Path(__file__))
+
+    with Context(
+        variant={},
+        iteration=0,
+        options={},
+        source=source,
+        variant_index=0,
+    ) as ctx:
+        with background(ctx.shell.__getattr__(sys.executable), "-c", COOPERATIVE) as bp:
```
xaru8145: Use python3 here like the other tests.

let's keep a consistent style.

xaru8145 (changes requested):

src/lambkin/cli.py
teresa-ortega: Solved in 81a50d2

src/lambkin/cli.py
teresa-ortega: Solved in 978c141

src/lambkin/cli.py
teresa-ortega: Solved in 5c88ba6

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in 0e507ae

src/lambkin/core/ctx/context.py
teresa-ortega: You're right, using the kill_and_remove_cgroup_tree function is much more robust. Solved in fdb36d5

test/integration/integration_test_cgroups_cleanup.py
teresa-ortega: Solved in fee42c5

test/integration/integration_test_cgroups_cleanup.py
teresa-ortega: Solved in 6ee9d8c

src/lambkin/core/process/cgroup.py
```diff
@@ -63,6 +63,23 @@ def find_delegated_cgroup() -> Path:
...
+    Path or None
+        The app.slice cgroup path if it exists and is writable, else None.
+    """
+    uid = os.getuid()
+    app_slice = Path(
+        f"/sys/fs/cgroup/user.slice/user-{uid}.slice/user@{uid}.service/app.slice"
+    )
+    if app_slice.exists() and os.access(app_slice, os.W_OK):
+        return app_slice
+    return None
```
xaru8145, line 84: Good! Maybe throw a message if not found.

src/lambkin/cli.py
```diff
@@ -109,8 +113,9 @@ def _handle_sigint(signum, frame):
     try:
         proc.wait()
     except KeyboardInterrupt:
-        kill_cgroup(run_cgroup)
+        kill_cgroup_tree(run_cgroup)
         remove_cgroup_tree(run_cgroup)
         sys.exit(130)
-
+    kill_cgroup_tree(run_cgroup)
```
xaru8145: Do we really need to kill here if everything worked fine? Just with removing should be fine no?

src/lambkin/cli.py
xaru8145: I think this needs to be updated too.

xaru8145 (changes requested):
Almost there

src/lambkin/cli.py
```diff
@@ -65,84 +73,49 @@ def format_help(self, ctx: click.Context, formatter: HelpFormatter) -> None:
...
         130  The benchmark was interrupted via Ctrl-C (SIGINT).
         N    Any other return code is propagated from the benchmark script.
     """
     # TODO(teresa-ortega): Handle concurrent runs, interrupted benchmarks, and re-runs
     # (e.g. detect an already active scope, support partial restarts).
     # To be addressed in phase 6.
-
-    cgroup_scope = f"lambkin-{script.stem}.scope"
+    parent = find_app_slice() or find_delegated_cgroup()
+    run_cgroup = make_cgroup(parent, f"lambkin-{script.stem}-{uuid.uuid4().hex[:8]}")
```
xaru8145, line 103: Printing the run_cgroup could be useful for debugging. This could be promoted to a log.debug in the logging PR.

examples/beluga/beluga_benchmark.py
```diff
@@ -33,26 +33,41 @@ def nominal(ctx):
...
-        ctx.source.path.parent / "beluga.launch.xml",
-        f"sensor_model:={ctx.variant.sensor_model}",
-        f"num_particles:={ctx.variant.num_particles}",
-    )
-    ctx.shell.ros2.bag.play("--clock", "-r", ctx.options.clock_rate)
-    ctx.shell.evo_ape.bag2("output.mcap", save_results="out.zip")
+    with lambkin.process.background(
+        ctx.shell.ros2.bag.record,
+        "--output",
+        "output.mcap",
```
xaru8145:
```suggestion
        "output",
```

src/lambkin/cli.py
```diff
@@ -65,84 +73,49 @@ def format_help(self, ctx: click.Context, formatter: HelpFormatter) -> None:
...
-                str(script),
-                *args,
-            ],
-        )
         proc.wait()
     except KeyboardInterrupt:
-        _stop_scope(cgroup_scope)
+        kill_cgroup_tree(run_cgroup)
+        remove_cgroup_tree(run_cgroup)
         sys.exit(130)
```
xaru8145, line 124: After doing **ctl+c**, the terminal behaves weird and does not respond well. That's because start_new_session=True detaches the child from the terminal's process group.  `stty sane` resets the terminal to a sane state, only after we check that we are running it in an interactive mode.
```suggestion
    except KeyboardInterrupt:
        kill_cgroup_tree(run_cgroup)
        remove_cgroup_tree(run_cgroup)
        if sys.stdin.isatty():
            os.system("stty sane")
        sys.exit(130)
```

Also, runnig ctl+c does not print any message to terminal and ends abruptly, which can be confusing for the user. I think it would be useful to print sth like "Interrupted, cleaning up benchmark processes..."

src/lambkin/cli.py
teresa-ortega: You're right, solved in 12f7f3a

src/lambkin/cli.py
teresa-ortega: Solved in 12f7f3a

examples/beluga/beluga_benchmark.py
teresa-ortega: Solved in 5b4cd9e

src/lambkin/cli.py
teresa-ortega: Solved in a13f91a

src/lambkin/core/process/cgroup.py
teresa-ortega: Solved in e489123

src/lambkin/cli.py
teresa-ortega: Solved in 713cb20

src/lambkin/core/process/cgroup.py
```diff
@@ -77,6 +77,11 @@ def find_app_slice() -> Path | None:
     )
     if app_slice.exists() and os.access(app_slice, os.W_OK):
         return app_slice
+    print(
+        f"[INFO] app.slice not available at {app_slice}, "
+        "falling back to delegated cgroup.",
```
xaru8145: "falling back to delegated cgroup." -> This part is not true. This is done in the CLI. Here we just return None, so let's just print the app.slice part not being available.

xaru8145 (changes requested):
Minor comment before approving

src/lambkin/core/process/cgroup.py
teresa-ortega: Solved in 82bd129

xaru8145 (approved):
LGTM. Great job! 🚀

PR #152, opened 2026-05-15
Phase 4: Add CLI entry point and native SDK options

teresa-ortega:
### Proposed changes

Exposes lambkin as a CLI command via a console script entry point. The main() function in cli.py wraps benchmark script execution inside a systemd-run --scope so that all child processes are placed in a dedicated cgroup v2 scope automatically, without requiring the user to do anything.

Adds sdk_options.py to define SDK-level CLI options like --dry-run that are always available on ctx.options without the user having to declare them via @lambkin.option. These are merged with user-defined options in _parse_options() inside benchmark.py.

Implemented the --help flag for lambkin, which now displays usage information and available options when the CLI is invoked with uv run lambkin --help.

#### How to test
1. Re-install the lambkin package:
```bash
uv sync
```
2. Run lambkin in --help mode:
```bash
uv run lambkin --help
```

3. Run the benchmark script:
```bash
```

#### Type of change

💥 No breaking changes

### Checklist

### Additional comments

N/A

### DIFF
--- a/src/lambkin/cli.py
+++ b/src/lambkin/cli.py
@@ -28,8 +28,60 @@ import subprocess
 import sys
 from pathlib import Path
 
+import click
+from click.formatting import HelpFormatter
 
-def main() -> None:
+
+class LambkinCommand(click.Command):
+    """Custom Click command that renders SDK and custom options separately."""
+
+    def format_help(self, ctx: click.Context, formatter: HelpFormatter) -> None:
+        """Write the full help text with SDK and custom options sections."""
+        formatter.write_paragraph()
+        formatter.write_text(
+            "Usage: lambkin [OPTIONS] SCRIPT [SDK_OPTIONS] [CUSTOM_OPTIONS]"
+        )
+        formatter.write_paragraph()
+        formatter.write_text(
+            "LAMBKIN is a benchmarking SDK for robotics applications. "
+            "It runs your benchmark script inside a systemd cgroup scope, "
+            "ensuring all child processes are tracked and cleaned up automatically."
+        )
+        formatter.write_paragraph()
+
+        with formatter.section("Options"):
+            formatter.write_dl([("--help", "Show this message and exit.")])
+
+        with formatter.section("SDK Options (always available)"):
+            formatter.write_dl(
+                [
+                    (
+                        "--dry-run",
+                        "Run the benchmark in dry-run mode: "
+                        "commands are logged but not executed.",
+                    ),
+                    (
+                        "--show-options",
+                        "List all SDK and custom options available "
+                        "for this benchmark script and exit.",
+                    ),
+                ]
+            )
+
+        with formatter.section("Custom Options (script-defined)"):
+            formatter.write_text(
+                "Options registered in your benchmark script via @lambkin.option.\n"
+                "\nRun 'lambkin SCRIPT --show-options' to list them."
+            )
+
+
+@click.command(
+    cls=LambkinCommand,
+    context_settings={"allow_extra_args": True, "ignore_unknown_options": True},
+)
+@click.argument("script", type=click.Path(exists=True, path_type=Path))
+@click.argument("args", nargs=-1, type=click.UNPROCESSED)
+def main(script: Path, args: tuple) -> None:
     """Launch a lambkin benchmark script inside a systemd cgroup scope.
 
     Re-executes ``script`` with the same Python interpreter, wrapped in
@@ -55,17 +107,6 @@ def main() -> None:
     # TODO(teresa-ortega): Handle concurrent runs, interrupted benchmarks, and re-runs
     # (e.g. detect an already active scope, support partial restarts).
     # To be addressed in phase 6.
-    if len(sys.argv) < 2:
-        print("Usage: lambkin <script.py> [args...]", file=sys.stderr)
-        sys.exit(1)
-
-    script = Path(sys.argv[1]).resolve()
-    if not script.exists():
-        print(f"Error: script not found: {script}", file=sys.stderr)
-        sys.exit(1)
-
-    args = sys.argv[2:]
-
     cgroup_scope = f"lambkin-{script.stem}.scope"
 
     try:
@@ -82,15 +123,13 @@ def main() -> None:
         )
     except KeyboardInterrupt:
         subprocess.run(
-            ["systemctl", "--user", "stop", cgroup_scope],
+            ["systemctl", "--user", "stop", "--wait", cgroup_scope],
             capture_output=True,
         )
         sys.exit(130)
-    except FileNotFoundError:
-        print(
-            "Error: 'systemd-run' not found. lambkin requires systemd.",
-            file=sys.stderr,
-        )
-        sys.exit(127)
+    except FileNotFoundError as err:
+        raise click.ClickException(
+            "'systemd-run' not found. lambkin requires systemd."
+        ) from err
 
     sys.exit(result.returncode)
--- a/src/lambkin/cli.py
+++ b/src/lambkin/cli.py
@@ -63,16 +63,21 @@ def main() -> None:
     args = sys.argv[2:]
 
     cgroup_scope = f"lambkin-{script.stem}.scope"
-
-    result = subprocess.run(
-        [
-            "systemd-run",
-            "--scope",
-            f"--unit={cgroup_scope}",
-            "--user",
-            sys.executable,
-            str(script),
-            *args,
-        ],
-    )
+    try:
+        result = subprocess.run(
+            [
+                "systemd-run",
+                "--scope",
+                f"--unit={cgroup_scope}",
+                "--user",
+                sys.executable,
+                str(script),
+                *args,
+            ],
+        )
+    except FileNotFoundError:
+        print(
+            "Error: 'systemd-run' not found. lambkin requires systemd.", file=sys.stderr
+        )
+        sys.exit(127)
     sys.exit(result.returncode)
--- a/src/lambkin/cli.py
+++ b/src/lambkin/cli.py
@@ -24,6 +24,7 @@ Typical usage::
     lambkin my_benchmark.py --clock-rate 50 --dry-run
 """
 
+import argparse
 import subprocess
 import sys
 from pathlib import Path
@@ -52,17 +53,42 @@ def main() -> None:
     SystemExit
         Always — propagates the child process return code.
     """
-    if len(sys.argv) < 2:
-        print("Usage: lambkin <script.py> [args...]", file=sys.stderr)
-        sys.exit(1)
+    parser = argparse.ArgumentParser(prog="lambkin")
+    parser.add_argument("script", type=str)
+    parser.add_argument(
+        "--overwrite",
+        action="store_true",
+        help="Stop any existing run of the same benchmark and re-launch.",
+    )
+    args, forwarded = parser.parse_known_args()
 
-    script = Path(sys.argv[1]).resolve()
+    script = Path(args.script).resolve()
     if not script.exists():
         print(f"Error: script not found: {script}", file=sys.stderr)
         sys.exit(1)
     args = sys.argv[2:]
 
     cgroup_scope = f"lambkin-{script.stem}.scope"
+    check = subprocess.run(
+        ["systemctl", "--user", "is-active", cgroup_scope],
+        capture_output=True,
+    )
+    scope_active = check.returncode == 0
+
+    if scope_active and not args.overwrite:
+        print(
+            f"Error: scope '{cgroup_scope}' is already active. "
+            "Use --overwrite to stop it and re-launch.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+
+    if scope_active and args.overwrite:
+        subprocess.run(
+            ["systemctl", "--user", "stop", cgroup_scope],
+            capture_output=True,
+        )
+
     try:
         result = subprocess.run(
             [
@@ -72,12 +98,20 @@ def main() -> None:
                 "--user",
                 sys.executable,
                 str(script),
-                *args,
+                *forwarded,
             ],
         )
+    except KeyboardInterrupt:
+        subprocess.run(
+            ["systemctl", "--user", "stop", cgroup_scope],
+            capture_output=True,
+        )
+        sys.exit(130)  # convención: 128 + SIGINT
     except FileNotFoundError:
         print(
-            "Error: 'systemd-run' not found. lambkin requires systemd.", file=sys.stderr
+            "Error: 'systemd-run' not found. lambkin requires systemd.",
+            file=sys.stderr,
         )
         sys.exit(127)
+
     sys.exit(result.returncode)
--- a/src/lambkin/cli.py
+++ b/src/lambkin/cli.py
@@ -66,7 +66,6 @@ def main() -> None:
     if not script.exists():
         print(f"Error: script not found: {script}", file=sys.stderr)
         sys.exit(1)
-    args = sys.argv[2:]
 
     cgroup_scope = f"lambkin-{script.stem}.scope"
     check = subprocess.run(
--- a/src/lambkin/cli.py
+++ b/src/lambkin/cli.py
@@ -31,6 +31,10 @@ from pathlib import Path
 import click
 from click.formatting import HelpFormatter
 
+from lambkin.common.exceptions import (
+    LambkinSystemdNotFoundError,
+    LambkinSystemdScopeTimeoutError,
+)
 from lambkin.sdk_options import SDK_OPTIONS
 
 
@@ -78,12 +82,11 @@ def _stop_scope(cgroup_scope: str) -> None:
             capture_output=True,
             timeout=30,
         )
-    except subprocess.TimeoutExpired:
-        click.echo(
-            f"Warning: timed out waiting for scope '{cgroup_scope}' to stop. "
-            "Some processes may still be running.",
-            err=True,
-        )
+    except subprocess.TimeoutExpired as err:
+        raise LambkinSystemdScopeTimeoutError(
+            f"Timed out waiting for scope '{cgroup_scope}' to stop. "
+            "Some processes may still be running."
+        ) from err
 
 
 @click.command(
@@ -140,7 +143,7 @@ def main(script: Path, args: tuple) -> None:
         _stop_scope(cgroup_scope)
         sys.exit(130)
     except FileNotFoundError as err:
-        raise click.ClickException(
+        raise LambkinSystemdNotFoundError(
             "'systemd-run' not found. lambkin requires systemd."
         ) from err
 
--- a/src/lambkin/cli.py
+++ b/src/lambkin/cli.py
@@ -56,7 +56,10 @@ def main() -> None:
         print("Usage: lambkin <script.py> [args...]", file=sys.stderr)
         sys.exit(1)
 
-    script = Path(sys.argv[1])
+    script = Path(sys.argv[1]).resolve()
+    if not script.exists():
+        print(f"Error: script not found: {script}", file=sys.stderr)
+        sys.exit(1)
     args = sys.argv[2:]
 
     cgroup_scope = f"lambkin-{script.stem}.scope"
--- a/src/lambkin/cli.py
+++ b/src/lambkin/cli.py
@@ -86,12 +86,11 @@ def main(script: Path, args: tuple) -> None:
     unchanged, so SDK and user-defined CLI options (e.g. ``--dry-run``,
     ``--clock-rate``) are passed through transparently.
 
-    Exits with the same return code as the child process.
-
-    Raises:
-    ------
-    SystemExit
-        Always — propagates the child process return code.
+    Exit codes:
+        0    The benchmark script completed successfully.
+        1    An error occurred (e.g. systemd-run not found).
+        130  The benchmark was interrupted via Ctrl-C (SIGINT).
+        N    Any other return code is propagated from the benchmark script.
     """
     # TODO(teresa-ortega): Handle concurrent runs, interrupted benchmarks, and re-runs
     # (e.g. detect an already active scope, support partial restarts).
--- a/src/lambkin/cli.py
+++ b/src/lambkin/cli.py
@@ -24,7 +24,6 @@ Typical usage::
     lambkin my_benchmark.py --clock-rate 50 --dry-run
 """
 
-import argparse
 import subprocess
 import sys
 from pathlib import Path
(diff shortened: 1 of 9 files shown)


src/lambkin/sdk_options.py
```diff
@@ -0,0 +1,38 @@
...
+# limitations under the License.
+
+"""SDK-level CLI options exposed natively by lambkin.
+
+Defines the set of CLI options that lambkin provides out of the box,
+without requiring the user to declare them via ``@lambkin.option``.
+These options are merged with any user-defined options by ``@benchmark``
+before parsing, so they are always available on ``ctx.options``.
+
+Typical usage::
```
xaru8145:
```suggestion
Typical usage:
```

src/lambkin/cli.py
```diff
@@ -0,0 +1,75 @@
...
+    ------
+    SystemExit
+        Always — propagates the child process return code.
+    """
+    if len(sys.argv) < 2:
+        print("Usage: lambkin <script.py> [args...]", file=sys.stderr)
+        sys.exit(1)
+
+    script = Path(sys.argv[1])
+    args = sys.argv[2:]
```
xaru8145: What happens if the script does not exist? Nothing is checked here and it is passed down to systemd, which I am not sure which error it will list. Please try and let me know. Otherwise considering a check before construction the cgroup scope.

src/lambkin/cli.py
```diff
@@ -0,0 +1,75 @@
...
+        [
+            "systemd-run",
+            "--scope",
+            f"--unit={cgroup_scope}",
+            "--user",
+            sys.executable,
+            str(script),
+            *args,
+        ],
+    )
```
xaru8145: What happens if systemd is not found? Who catches the error? Is it graceful and clear? Should we do sth here at the entry point?

src/lambkin/cli.py
```diff
@@ -0,0 +1,75 @@
...
+        Always — propagates the child process return code.
+    """
+    if len(sys.argv) < 2:
+        print("Usage: lambkin <script.py> [args...]", file=sys.stderr)
+        sys.exit(1)
+
+    script = Path(sys.argv[1])
+    args = sys.argv[2:]
+
+    cgroup_scope = f"lambkin-{script.stem}.scope"
```
xaru8145, line 120: If two benchmarks with the same run concurrently (or a previous one didn't clean up), systemd will could the scope name with a non-obvious error.

First, let's think what should happen if we try to run the same benchmark script twice. Maybe detect that it already completed and fail. If it was half way, resume it (partial restart) but did it clean correctlty the process groups? What if it already run correctly but we want to re run it, should we expose a flag like` --re-run` or `--overwrite`? Speaking out loud here to raise a discussion.

Have you tested interruptions in the process leave a dangling cgroup at the benchmark level per context?

src/lambkin/sdk_options.py
```diff
@@ -0,0 +1,38 @@
...
+
+    # ctx.options.dry_run is always available, no @lambkin.option needed
+    def my_benchmark(ctx):
+        ctx.options.dry_run
+"""
+
+import click
+
+SDK_OPTIONS: list[click.Option] = [
```
xaru8145: `SDK_OPTIONS` is a plain list. Nothing prevents user code from accidentally mutating it (e.g., `SDK_OPTIONS.append(...)`). Consider promoting it to a tuple, which is inmutable and then create a test to ensure we can't append any element.

When you fuse it in the benchmark with `user_options`, you'll probably need to convert it back to a list. But you can do it in place in that same operations.

src/lambkin/sdk_options.py
```diff
@@ -0,0 +1,38 @@
...
+    click.Option(
+        ["--dry-run"],
+        is_flag=True,
+        default=False,
+        help="Print commands without executing them.",
```
xaru8145: This description is a bit vague. I suggest sth more like: "Run the benchmark in dry-run mode: commands are logged but not executed."

test/core/test_benchmark.py
xaru8145, line 1: Not sure if there is a test already but we should make sure that `ctx.options.dry_run` is actually True when `--dry-run` is passed. Maybe a test for the context rather than the benchmark.

src/lambkin/sdk_options.py
```diff
@@ -0,0 +1,38 @@
...
+without requiring the user to declare them via ``@lambkin.option``.
+These options are merged with any user-defined options by ``@benchmark``
+before parsing, so they are always available on ``ctx.options``.
+
+Typical usage::
+
+    # ctx.options.dry_run is always available, no @lambkin.option needed
+    def my_benchmark(ctx):
+        ctx.options.dry_run
+"""
```
xaru8145, line 35: If a user somehow declares a `--dry-run` option manually via `@lambkin.option`, Click will raise a duplicate parameter error. This is expected but we should document explicitly here that _Option names here are reserved. Declaring them via @lambkin.option in user benchmarks will raise a Click error at parse time_.

src/lambkin/cli.py
```diff
@@ -0,0 +1,75 @@
...
+    Raises:
+    ------
+    SystemExit
+        Always — propagates the child process return code.
+    """
+    if len(sys.argv) < 2:
+        print("Usage: lambkin <script.py> [args...]", file=sys.stderr)
+        sys.exit(1)
+
+    script = Path(sys.argv[1])
```
xaru8145: I think this does not resolve relative paths. What if we run `lambkin ../other/benchmark.py`? Will inspect.getfile called in the Benchmark work correctly? If not, consider using [resolve](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve)
```suggestion
    script = Path(sys.argv[1]).resolve()
```
And add a test to the cli in which this path resolution is asserted correctly.

xaru8145 (changes requested):

src/lambkin/cli.py
teresa-ortega: Solved in 2738a6a

src/lambkin/cli.py
teresa-ortega: Solved in 26ee3ef

test/core/test_benchmark.py
```diff
@@ -41,7 +41,7 @@ def fn(ctx):
         pass
 
     result = _parse_options(fn, [])
-    assert result == {}
+    assert result == {"dry_run": False}
```
marcoshuck: Sounds like this should be/have a sensible default and avoid asserting it throughout the code

src/lambkin/cli.py
```diff
@@ -0,0 +1,117 @@
...
+"""Entry point for the lambkin CLI.
+
+Exposes the ``lambkin`` command, registered as a console script in
+``pyproject.toml``. When invoked, it re-executes the given benchmark
+script under a transient systemd scope so that all child processes
+are placed in a dedicated cgroup v2 hierarchy automatically.
+
+Typical usage::
+
+    lambkin my_benchmark.py --clock-rate 50 --dry-run
```
marcoshuck, line 23: is there a utility tool like [cobra](https://github.com/spf13/cobra) for clis in Python? just exploring the idea, I'd suggest we make this decision sooner rather than later, not something that needs to be addressed right now though.

marcoshuck: looks like `click` is the actual library you're using, nice :)

src/lambkin/cli.py
teresa-ortega: Solved in 2738a6a and 61c8275

src/lambkin/cli.py
teresa-ortega: Solved in ebb7b96 and 61c8275

src/lambkin/sdk_options.py
teresa-ortega: Solved in 369b43a

src/lambkin/sdk_options.py
teresa-ortega: Solved in 369b43a

src/lambkin/sdk_options.py
teresa-ortega: Solved in 369b43a

src/lambkin/sdk_options.py
teresa-ortega: Solved in 6ba2b9e

test/core/test_benchmark.py
teresa-ortega: Added two tests in `test/core/test_benchmark.py`: 
`test_parse_options_includes_dry_run_by_default` to verify `dry_run` is always present with a default of `False`, and `test_parse_options_dry_run_can_be_set_via_cli` to verify it returns True when `--dry-run` is passed via CLI.

test/core/test_benchmark.py
xaru8145: It does have a default here https://github.com/Ekumen-OS/lambkin/blob/6ba2b9ed8dd5a70a3a8032f61c60a6ec4e63ffe6/src/lambkin/sdk_options.py#L43

But it is true that this is hard to maintain as it is defined in two places. We could define the sdk options default value for dry mode as a constant so that we can use it here and there.

src/lambkin/cli.py
xaru8145: Yes, LAMBKIN-ng is modeled after click! Thanks for the suggestion though

src/lambkin/cli.py
```diff
@@ -0,0 +1,116 @@
...
+                str(script),
+                *forwarded,
+            ],
+        )
+    except KeyboardInterrupt:
+        subprocess.run(
+            ["systemctl", "--user", "stop", cgroup_scope],
+            capture_output=True,
+        )
+        sys.exit(130)  # convención: 128 + SIGINT
```
xaru8145: spanish comment

src/lambkin/cli.py
```diff
@@ -0,0 +1,116 @@
...
+``pyproject.toml``. When invoked, it re-executes the given benchmark
+script under a transient systemd scope so that all child processes
+are placed in a dedicated cgroup v2 hierarchy automatically.
+
+Typical usage::
+
+    lambkin my_benchmark.py --clock-rate 50 --dry-run
+"""
+
+import argparse
```
xaru8145: Why are we using argparse here? LAMBKIN is modeled after click so I think this is incosistent.

src/lambkin/cli.py
xaru8145: @teresa-ortega My comment was trying to open a discussion and raise some questions. I would have expected answers to those questions and some back and forth before you implemented anything. Let's try to be more efficient here.

src/lambkin/cli.py
```diff
@@ -0,0 +1,116 @@
...
+    script = Path(args.script).resolve()
+    if not script.exists():
+        print(f"Error: script not found: {script}", file=sys.stderr)
+        sys.exit(1)
+
+    cgroup_scope = f"lambkin-{script.stem}.scope"
+    check = subprocess.run(
+        ["systemctl", "--user", "is-active", cgroup_scope],
+        capture_output=True,
+    )
```
xaru8145: This is unguarded. If there is no systemd this will throw.

src/lambkin/cli.py
xaru8145: I am going to push back a bit against this implementation. I think it is a bit brittle and we need to think the possible cases and how to deal with them. Please revert back this, add a todo comment and let's have a simple version that we can improve and robustify in the near future. I think this would make sense to be introduced in phase 6 along with partial restarts.

xaru8145 (changes requested):

src/lambkin/cli.py
teresa-ortega: Sorry for jumping ahead without discussing it first. I should have answered your questions before implementing anything. I will revert the implementation and leave a TODO comment so we can revisit this properly in phase 6.

teresa-ortega: Done in fae5fb6

teresa-ortega: > If two benchmarks with the same run concurrently (or a previous one didn't clean up), systemd will could the scope name with a non-obvious error.
> 
> First, let's think what should happen if we try to run the same benchmark script twice. Maybe detect that it already completed and fail. If it was half way, resume it (partial restart) but did it clean correctlty the process groups? What if it already run correctly but we want to re run it, should we expose a flag like` --re-run` or `--overwrite`? Speaking out loud here to raise a discussion.
> 
> Have you tested interruptions in the process leave a dangling cgroup at the benchmark level per context?

If the scope is already active it could mean a concurrent run or a crashed one. For the concurrent case the simplest option is probably to fail with a clear error message. For a dangling scope after a crash, one option could be an --overwrite flag that cleans it up and re-launches, though we could also try to detect automatically whether the scope is genuinely active or a leftover from a previous run.

src/lambkin/cli.py
teresa-ortega: This is due to the overwrite implementation, which has already been reverted in fae5fb6

src/lambkin/cli.py
teresa-ortega: This is due to the overwrite implementation, which has already been reverted in fae5fb6

src/lambkin/cli.py
teresa-ortega: Solved in fae5fb6

test/core/test_benchmark.py
```diff
@@ -177,3 +177,24 @@ def test_benchmark_empty_variants_raises_error(tmp_path):
...
+
+def test_parse_options_includes_dry_run_by_default():
+    """_parse_options always includes dry_run even when no user options are declared."""
+
+    def fn(ctx):
+        pass
+
+    result = _parse_options(fn, [])
+    assert "dry_run" in result
+    assert result["dry_run"] is False
```
xaru8145:
```suggestion
    assert result["dry_run"] is defaults.DRY_RUN
```

test/core/test_benchmark.py
xaru8145:
```suggestion
    assert result == {"dry_run": defaults.DRY_RUN}
```

test/core/test_benchmark.py
```diff
@@ -53,7 +53,7 @@ def fn(ctx):
         pass
 
     result = _parse_options(fn, [])
-    assert result == {"clock_rate": 100.0, "sensor_topic": "/scan"}
+    assert result == {"dry_run": False, "clock_rate": 100.0, "sensor_topic": "/scan"}
```
xaru8145:
```suggestion
    assert result == {"dry_run": defaults.DRY_RUN, "clock_rate": 100.0, "sensor_topic": "/scan"}
```

test/core/test_benchmark.py
```diff
@@ -65,7 +65,7 @@ def fn(ctx):
         pass
 
     result = _parse_options(fn, ["--clock-rate", "50.0"])
-    assert result == {"clock_rate": 50.0, "sensor_topic": "/scan"}
+    assert result == {"dry_run": False, "clock_rate": 50.0, "sensor_topic": "/scan"}
```
xaru8145:
```suggestion
    assert result == {"dry_run": defaults.DRY_RUN, "clock_rate": 50.0, "sensor_topic": "/scan"}
```

src/lambkin/cli.py
```diff
@@ -0,0 +1,136 @@
...
+
+    Raises:
+    ------
+    SystemExit
+        Always — propagates the child process return code.
+    """
+    # TODO(teresa-ortega): Handle concurrent runs, interrupted benchmarks, and re-runs
+    # (e.g. detect an already active scope, support partial restarts).
+    # To be addressed in phase 6.
+    script = script.resolve()
```
xaru8145: Is this needed now that click handles the script path?

src/lambkin/cli.py
```diff
@@ -0,0 +1,136 @@
...
+                        "Run the benchmark in dry-run mode: "
+                        "commands are logged but not executed.",
+                    ),
+                    (
+                        "--show-options",
+                        "List all SDK and custom options available "
+                        "for this benchmark script and exit.",
+                    ),
+                ]
+            )
```
xaru8145:
```suggestion
        with formatter.section("SDK Options (always available)"):
            formatter.write_dl([
                (opt.opts[0], opt.help or "")
                for opt in SDK_OPTIONS
            ])
```

src/lambkin/cli.py
```diff
@@ -0,0 +1,136 @@
...
+
+        with formatter.section("Custom Options (script-defined)"):
+            formatter.write_text(
+                "Options registered in your benchmark script via @lambkin.option.\n"
+                "\nRun 'lambkin SCRIPT --show-options' to list them."
+            )
```
xaru8145, line 64: This shows with an extra new line:
```bash
Custom Options (script-defined):
  Options registered in your benchmark script via @lambkin.option.

  Run 'lambkin SCRIPT --show-options' to list them.
```

Better like this:
```suggestion
            formatter.write_text(
                "Options registered in your benchmark script via @lambkin.option."
            )
            formatter.write_text("Run 'lambkin SCRIPT --show-options' to list them.")
```

test/test_cli.py
xaru8145, line 1: Consider adding a test where help output contains the expected sections.

src/lambkin/cli.py
```diff
@@ -0,0 +1,136 @@
...
+    process tree can be inspected and killed cleanly without leaving
+    orphaned processes behind.
+
+    The cgroup scope is named ``lambkin-<stem>.scope``, where ``<stem>``
+    is the filename of the script without its extension.  All arguments
+    that follow the script path are forwarded to the child process
+    unchanged, so SDK and user-defined CLI options (e.g. ``--dry-run``,
+    ``--clock-rate``) are passed through transparently.
+
+    Exits with the same return code as the child process.
```
xaru8145: If you want and think can be useful, you can add more details on the codes we can return. Example:
```py
"""
Exit codes:
        0    The benchmark script completed successfully.
        130  The benchmark was interrupted via Ctrl-C (SIGINT).
        N    Any other return code is propagated from the benchmark script.
"""
```

src/lambkin/cli.py
```diff
@@ -0,0 +1,136 @@
...
+
+    Raises:
+    ------
+    SystemExit
+        Always — propagates the child process return code.
```
xaru8145: We are catching the exceptions, so system exit is the return code, not an exception raise. The only exception we throw is a click exception when systemd is not found.

src/lambkin/cli.py
```diff
@@ -0,0 +1,136 @@
...
+                sys.executable,
+                str(script),
+                *args,
+            ],
+        )
+    except KeyboardInterrupt:
+        subprocess.run(
+            ["systemctl", "--user", "stop", "--wait", cgroup_scope],
+            capture_output=True,
+        )
```
xaru8145: We can consider adding a generous timeout just in case there is any blocked process that is not properly terminating and log accordingly.

src/lambkin/cli.py
```diff
@@ -0,0 +1,136 @@
...
+                "systemd-run",
+                "--scope",
+                f"--unit={cgroup_scope}",
+                "--user",
+                sys.executable,
+                str(script),
+                *args,
+            ],
+        )
+    except KeyboardInterrupt:
```
xaru8145, line 139: `KeyboardInterrupt` catch is unreliable due to process group signal delivery.

Ctrl-C sends `SIGINT` to the entire foreground process group simultaneously, both Python and `systemd-run` receive it at the same time. If `systemd-run` exits before Python's `except KeyboardInterrupt` executes, the scope may already be in a `deactivating` state and `systemctl stop --wait` may return before all child processes are actually dead, defeating the cleanup guarantee.

Claude suggested using `subprocess.Popen` with an explicit `signal.signal(signal.SIGINT, ...)` handler instead. Python intercepts `SIGINT` before `systemd-run` sees it, ensuring `systemctl stop --wait` always runs against a live scope.

xaru8145 (changes requested):

test/core/test_benchmark.py
teresa-ortega: Solved in fad0acb

test/core/test_benchmark.py
teresa-ortega: Solved in fad0acb

test/core/test_benchmark.py
teresa-ortega: Solved in fad0acb

test/core/test_benchmark.py
teresa-ortega: Solved in fad0acb

src/lambkin/cli.py
teresa-ortega: Solved in 21807e0

src/lambkin/cli.py
teresa-ortega: Solved in 21807e0

xaru8145 (commented):
Make sure to update the beluga benchmark example to not register the dry run flag on the script, as it is now part of SDK options.

https://github.com/Ekumen-OS/lambkin/blob/e77cb2dfd6fc3aa2cb0350fef9f8f2963034aa91/examples/beluga/beluga_benchmark.py#L27

src/lambkin/cli.py
teresa-ortega: You're right, it's redundant since the click already does it. This change affected the tests, but it's fixed in 57d55c3

test/test_cli.py
teresa-ortega: That's a really good idea. I added it in fa4d0c1

src/lambkin/cli.py
teresa-ortega: Added it in 02e856a

src/lambkin/cli.py
teresa-ortega: Okay, solved in 02e856a

src/lambkin/cli.py
teresa-ortega: I have implemented the suggestion using Popen with an explicit signal.SIGINT handler so Python intercepts the signal before systemd-run receives it, ensuring the scope is still alive when systemctl stop runs. Implemented in 5055086

src/lambkin/cli.py
teresa-ortega: I added a 30 second timeout to the stop call and log a warning if it expires in 5055086

src/lambkin/cli.py
```diff
@@ -64,6 +65,28 @@ def format_help(self, ctx: click.Context, formatter: HelpFormatter) -> None:
...
+            ["systemctl", "--user", "stop", cgroup_scope],
+            capture_output=True,
+            timeout=30,
+        )
+    except subprocess.TimeoutExpired:
+        click.echo(
+            f"Warning: timed out waiting for scope '{cgroup_scope}' to stop. "
+            "Some processes may still be running.",
+            err=True,
+        )
```
xaru8145: Why use click to log here? I would have thought of a print statement and once logging is in place we can use `logging` library.

test/core/test_benchmark.py
xaru8145: Done in fad0acb7a36d27a28de9b66fd5f7bc6bce157b5a

src/lambkin/cli.py
xaru8145: Clarified in 1-1 conversation. This throws an exception but it can be confusing to see a click exception. Better to use our own exception for that. Same for the other click exception that is thrown in this same file.

teresa-ortega: Solved in 894f5ad

src/lambkin/common/exceptions.py
```diff
@@ -17,3 +17,11 @@
...
 catch SDK-specific failures with a single except clause.
 """
+
+
+class LambkinSystemdNotFoundError(Exception):
+    """Raised when systemd-run is not found on the system."""
+
+
+class LambkinSystemdScopeTimeoutError(Exception):
+    """Raised when a systemd scope fails to stop within the timeout period."""
```
xaru8145: According to the documentation it should be like this:

```suggestion
class LambkinError(Exception):
    """Base class for all lambkin SDK errors."""

class LambkinSystemdNotFoundError(LambkinError):
    """Raised when systemd-run is not found on the system."""

class LambkinSystemdScopeTimeoutError(LambkinError):
    """Raised when a systemd scope fails to stop within the timeout period."""
```

src/lambkin/common/__init__.py
```diff
@@ -19,9 +19,12 @@
 
 from lambkin.common import defaults
 
+from .exceptions import LambkinSystemdNotFoundError, LambkinSystemdScopeTimeoutError
```
xaru8145:
```suggestion
from lambkin.common import exceptions
```

src/lambkin/common/__init__.py
```diff
@@ -19,9 +19,12 @@
...
 from .named_product import named_product
 
 __all__ = [
     "defaults",
     "named_product",
+    "LambkinSystemdNotFoundError",
+    "LambkinSystemdScopeTimeoutError",
```
xaru8145:
```suggestion
    "exceptions",
```

src/lambkin/cli.py
```diff
@@ -0,0 +1,150 @@
...
+import sys
+from pathlib import Path
+
+import click
+from click.formatting import HelpFormatter
+
+from lambkin.common.exceptions import (
+    LambkinSystemdNotFoundError,
+    LambkinSystemdScopeTimeoutError,
+)
```
xaru8145:
```suggestion
from lambkin.common import exceptions
```

xaru8145 (commented):
This way adding a new exception to `exceptions.py` doesn't require touching __init__.py.

src/lambkin/common/exceptions.py
teresa-ortega: Solved in 22bd26a

src/lambkin/common/__init__.py
teresa-ortega: Solved in 22bd26a

src/lambkin/cli.py
teresa-ortega: Solved in 22bd26a

src/lambkin/common/__init__.py
teresa-ortega: Solved in 22bd26a

xaru8145 (approved):

PR #153, opened 2026-05-20
Adds --show-options to the lambkin CLI

teresa-ortega:
### Proposed changes

When passed after a script, it imports the script without executing it and prints all options registered via `@lambkin.option`, then exits. 

Builds on top of :
* #152 

#### How to test
1. Re-install the lambkin package:
```bash
uv sync
```

2. Run the benchmark script with --show-options
```
uv run lambkin examples/beluga/beluga_benchmark.py --show-options
```

#### Type of change

💥 No breaking changes

### Checklist

### Additional comments

N/A

### DIFF
--- a/src/lambkin/core/decorators/benchmark.py
+++ b/src/lambkin/core/decorators/benchmark.py
@@ -32,6 +32,7 @@ import inspect
 import sys
 
 import click
+from click.formatting import HelpFormatter
 
 from lambkin.core.ctx.context import Context
 from lambkin.core.ctx.source import Source
@@ -40,9 +41,7 @@ from lambkin.sdk_options import SDK_OPTIONS
 
 
 def _show_options(fn) -> None:
-    """Print all options registered via @lambkin.option on fn and exit."""
-    from click.formatting import HelpFormatter
-
+    """Print all options registered via @lambkin.option on fn."""
     user_options = getattr(fn, "__lambkin_options__", [])
     formatter = HelpFormatter()
     if not user_options:
--- a/src/lambkin/core/decorators/benchmark.py
+++ b/src/lambkin/core/decorators/benchmark.py
@@ -44,10 +44,12 @@ def _show_options(fn) -> None:
     """Print all options registered via @lambkin.option on fn."""
     user_options = getattr(fn, "__lambkin_options__", [])
     formatter = HelpFormatter()
-    if not user_options:
-        formatter.write_text("No options registered in this script.")
-    else:
-        with formatter.section("Custom Options"):
+    with formatter.section("SDK Options"):
+        formatter.write_dl([(opt.opts[0], opt.help or "") for opt in SDK_OPTIONS])
+    with formatter.section("Custom Options"):
+        if not user_options:
+            formatter.write_text("No options registered in this script.")
+        else:
             formatter.write_dl(
                 [
                     (
--- a/src/lambkin/core/decorators/benchmark.py
+++ b/src/lambkin/core/decorators/benchmark.py
@@ -39,6 +39,18 @@ from lambkin.core.decorators.input import InputRegistry
 from lambkin.sdk_options import SDK_OPTIONS
 
 
+def _show_options(fn) -> None:
+    """Print all options registered via @lambkin.option on fn and exit."""
+    user_options = getattr(fn, "__lambkin_options__", [])
+    if not user_options:
+        print("No options registered in this script.")
+    else:
+        print("Custom Options:")
+        for opt in user_options:
+            default = f"  [default: {opt.default}]" if opt.default is not None else ""
+            print(f"  {opt.opts[0]:<30} {opt.help or ''}{default}")
+
+
 def _parse_options(fn, cli_args):
     """Parse CLI options from fn.__lambkin_options__ and return a dict."""
     user_options = getattr(fn, "__lambkin_options__", [])
@@ -85,6 +97,9 @@ def benchmark(variants, num_iterations):
         def wrapper(args=None, output_dir=None):
             cli_args = sys.argv[1:] if args is None else args
             options = _parse_options(fn, cli_args)
+            if options.get("show_options"):
+                _show_options(fn)
+                sys.exit(0)
             source = Source(path=inspect.getfile(fn))
             # The base context creates a directory for variant 1 / iteration 1
             # containing a metadata file with the information available at this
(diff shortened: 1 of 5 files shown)


xaru8145 (changes requested):
The flag is intercepted at the CLI level as a raw string match rather than going through Click's option parsing machinery, and it won't appear in lambkin `--help` under SDK Options. I am not convinced this is the implementation we want.

To complete this properly:

- Add `--show-options` to `SDK_OPTIONS` in `sdk_options.py`
- Handle it inside `_parse_options` in `benchmark.py` rather than intercepting it in cli.py, the CLI should forward it transparently like any other SDK option. Why? Because running the full script to discover options is fragile and may trigger side effects. Instead, drive option discovery through the SDK's own registration mechanism.

src/lambkin/sdk_options.py
```diff
@@ -26,12 +26,18 @@
...
+        - ``--show-options``
 
-Typical usage:
+Typical usage::
 
     # ctx.options.dry_run is always available, no @lambkin.option needed
     def my_benchmark(ctx):
-        ctx.options.dry_run
+        if ctx.options.dry_run:
+            ...
```
xaru8145: I think dry run is not supposed to work like this, so the typical usage comment can be misleading. Would remove it.

src/lambkin/sdk_options.py
```diff
@@ -46,4 +52,11 @@ def my_benchmark(ctx):
         default=defaults.DRY_RUN,
         help="Run the benchmark in dry-run mode: commands are logged but not executed.",
     ),
+    click.Option(
+        ["--show-options"],
+        is_flag=True,
+        show_default=True,
```
xaru8145: I think `show_default` this is not needed as this is a show and exit flag.

src/lambkin/core/decorators/benchmark.py
```diff
@@ -39,6 +39,18 @@
...
+def _show_options(fn) -> None:
+    """Print all options registered via @lambkin.option on fn and exit."""
+    user_options = getattr(fn, "__lambkin_options__", [])
+    if not user_options:
+        print("No options registered in this script.")
+    else:
+        print("Custom Options:")
+        for opt in user_options:
+            default = f"  [default: {opt.default}]" if opt.default is not None else ""
+            print(f"  {opt.opts[0]:<30} {opt.help or ''}{default}")
```
xaru8145: Why not use click's `HelpFormatter` for consistency with `lambkin --help` rather than raw prints?

test/test_cli.py
```diff
@@ -85,3 +87,39 @@ def test_help_output_contains_expected_sections():
...
+
+
+def test_show_options_does_not_reach_systemd(dummy_script):
+    """--show-options is forwarded to subprocess like any other SDK option."""
+    runner = CliRunner()
+    with patch("subprocess.Popen") as mock_popen:
+        mock_popen.return_value = MagicMock(returncode=0)
+        runner.invoke(main, [str(dummy_script), "--show-options"])
+        cmd = mock_popen.call_args[0][0]
+        assert "--show-options" in cmd
```
xaru8145: `_show_options` is a private function (underscore prefix) from benchmark. Importing it and testing it doe snot feel right and directly couples the test to the internal structure. Test it through the public interface instead: invoke the benchmark wrapper with --show-options and assert on stdout.

xaru8145 (changes requested):

src/lambkin/sdk_options.py
teresa-ortega: Solved in af3059e

src/lambkin/sdk_options.py
teresa-ortega: Solved in 8c5cab5

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in d63c7d5

test/test_cli.py
teresa-ortega: Addressed. Moved the tests to `test_benchmark.py` and test through the public `@benchmark` wrapper instead of calling `_show_options` directly. Used `pytest.raises(SystemExit)` to assert the exit code and `capsys` to capture the output.
Implemented in e9395a4

src/lambkin/core/decorators/benchmark.py
```diff
@@ -39,6 +39,33 @@
 from lambkin.sdk_options import SDK_OPTIONS
 
 
+def _show_options(fn) -> None:
+    """Print all options registered via @lambkin.option on fn and exit."""
+    from click.formatting import HelpFormatter
```
xaru8145: Move to the top with the rest of imports.

src/lambkin/core/decorators/benchmark.py
```diff
@@ -39,6 +39,33 @@
 from lambkin.sdk_options import SDK_OPTIONS
 
 
+def _show_options(fn) -> None:
+    """Print all options registered via @lambkin.option on fn and exit."""
```
xaru8145: You document here that you exit but you actually never exit. The exit happens in the wrapper. Be consistent.

src/lambkin/core/decorators/benchmark.py
```diff
@@ -39,6 +39,33 @@
...
+                        + (
+                            f"  [default: {opt.default}]"
+                            if opt.default is not None
+                            else ""
+                        ),
+                    )
+                    for opt in user_options
+                ]
+            )
+    click.echo(formatter.getvalue(), nl=False)
```
xaru8145: What do you think about printing all options (sdk and user). Otherwise the user might have to switch between --help and --show-options to get a full picture of the options they have for their script. I envisioned this as a unified way to see all flags that can be passed to the script, including the ones that come by default in the sdk. What do you think?

xaru8145 (changes requested):

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in c4673db

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in c4673db

src/lambkin/core/decorators/benchmark.py
teresa-ortega: I think that's a great idea. This way, the user also has access from show-options to the options they can enter after the script. It's implemented in 330eb67

xaru8145 (approved):
LGTM

PR #154, opened 2026-05-25
Phase 4: Logging Implementation

teresa-ortega:
### Proposed changes

Proposed changes
This PR implements process output logging for all commands dispatched through ShellProxy, as described in the logging decision document.

* ``CommandProxy.__call__`` has been switched from ``subprocess.run`` to ``subprocess.Popen`` to gain full control over stdout and stderr. The output destination is resolved following a three-level priority chain: the CLI ``--log-output`` flag overrides everything, then a per-call log_output kwarg, and finally the default ``"file"``.
* Log files are written to the iteration directory and named after the command tokens, without including positional arguments in the name. If the same command is launched more than once within a single iteration, a numeric suffix is appended to avoid collisions.
* The same output routing applies to background processes via ``lambkin.process.background``. The ``background()`` function intercepts the ``log_output`` kwarg and opens log files through the proxy before the process starts, so that foreground and background commands behave consistently.
* A specialisation registry has been added to ``CommandProxy`` to allow ROS-specific proxies to be plugged in without introducing a circular import. ``RosLaunchCommand`` uses this mechanism to inject ``ROS_LOG_DIR`` into the child process environment, so that ROS node logs land in the iteration directory alongside all other benchmark artefacts. The specialisation is registered at import time via ``core/shell/__init__.py``.
* A ``--log-level ``SDK option has been added to control the verbosity of lambkin's own output. It is parsed by both ``cli.py ``and ``benchmark.py``, and forwarded from the CLI to the script so both processes respect the same flag. The lambkin logger is fully isolated from the root logger ,user scripts can configure their own logging without any interference.

Test coverage has been added for:

* Log output priority resolution
* Log file naming and collision avoidance
* The specialisation registry returning ``RosLaunchCommand`` for ``ros2 launch`` 
* ``ROS_LOG_DIR`` injection
* Log file creation in file mode for both foreground and background commands
* The ``--log-level`` option parsin

#### How to test
Run the Beluga benchmark with the default output mode:
```bash
uv sync
```
Check that stdout and stderr log files have been created in the iteration directory:
```bash
ls examples/beluga/results/var_1/iter_1/
```
Expected: ros2_launch.stdout.log and ros2_launch.stderr.log present
Rerun with console mode and confirm that output reaches the terminal and no log files are created:
```bash
uv run lambkin examples/beluga/beluga_benchmark.py --log-output console
```
Verify log level control:

Expected output per log level:

* **debug** — [DRY RUN BG] and [DRY RUN] messages visible for all commands, plus internal SDK messages such as the cgroup path.
* **info** — [DRY RUN BG] and [DRY RUN] messages visible.
* **warning** — no lambkin output.

```bash
uv run lambkin examples/beluga/beluga_benchmark.py --log-level <level>
```

#### Type of change

💥 No breaking changes

### Checklist

### Additional comments

N/A


test/core/test_shell.py
```diff
@@ -156,3 +156,41 @@ def test_path_with_spaces_no_shell_injection(shell, tmp_path):
...
+
+def test_log_output_per_call_overrides_default(shell):
+    """Per-call log_output overrides the default."""
+    assert shell.echo._resolve_log_output("console") == "console"
+
+
+def test_log_output_cli_overrides_per_call(tmp_path):
+    """CLI log_output overrides per-call value."""
+    s = ShellProxy(dry_run=False, cwd=tmp_path, log_output="both")
+    assert s.echo._resolve_log_output("console") == "both"
```
xaru8145: The test uses `both` as log output but this option is not supported, as mentioned earlier.

src/lambkin/sdk_options.py
```diff
@@ -51,4 +51,11 @@
         default=False,
         help="List all options registered via @lambkin.option.",
     ),
+    click.Option(
+        ["--log-output"],
+        type=click.Choice(["file", "console"]),
+        show_default=True,
+        default=None,
+        help="Where to route process output: file, console, or both.",
```
xaru8145: "both" mode is mentioned in the PR description and the documentation of `log_output` in `sdk_options.py`, but this click option only accepts ["file", "console"] as valid choices. 

Make everything consistent. Is it supported or not? Let's try to not put much effort into it so if it is not there, do not implement it and just leave a TODO comment.

src/lambkin/core/shell/proxy.py
```diff
@@ -142,17 +254,28 @@ def __call__(self, *args: Any, **kwargs: Any) -> subprocess.CompletedProcess:
...
+                proc = self._make_popen(argv, None, None)
+                stdout, stderr = None, None
+
+            proc.wait()
+            if stdout:
+                stdout.close()
+            if stderr:
+                stderr.close()
+            if proc.returncode != 0:
+                raise subprocess.CalledProcessError(proc.returncode, argv)
```
xaru8145, line 311: Why do we raise here?

src/lambkin/core/shell/proxy.py
```diff
@@ -92,7 +121,89 @@ def __getattr__(self, name: str) -> _CommandProxy:
...
+
+        Returns:
+            A tuple of (stdout_file, stderr_file) open for writing.
+        """
+        if self._cwd is None:
+            return None, None
+        base = self._log_base(argv)
+        out = open(self._cwd / f"{base}.stdout.log", "w")
+        err = open(self._cwd / f"{base}.stderr.log", "w")
+        return out, err
```
xaru8145: `_open_streams` uses bare `open()` without a context manager or explicit close guarantee. The files are closed manually in `__call__`, but only in the happy path. If `_make_popen` raises, the file handles leak.

src/lambkin/core/shell/proxy.py
```diff
@@ -92,7 +121,89 @@ def __getattr__(self, name: str) -> _CommandProxy:
...
+        Returns:
+            A unique base name string, e.g. 'ros2_launch' or 'ros2_launch_1'.
+        """
+        base = "_".join(
+            arg for arg in argv if not arg.startswith("/") and ":=" not in arg
+        )
+        count = self._call_counts.get(base, 0)
+        self._call_counts[base] = count + 1
+        suffix = f"_{count}" if count > 0 else ""
+        return f"{base}{suffix}"
```
xaru8145: `_log_base` both derives the name and increments the counter, so calling it more than once per dispatch (e.g. in a debug log) would silently corrupt the suffix logic. Worth splitting into a pure name derivation and a separate method that owns the counter increment.

test/core/test_shell.py
```diff
@@ -156,3 +156,41 @@ def test_path_with_spaces_no_shell_injection(shell, tmp_path):
...
+
+
+def test_log_base_appends_suffix_on_collision(shell):
+    """_log_base appends numeric suffix when same command launched twice."""
+    proxy = shell.ros2
+    assert proxy._log_base(["ros2", "launch"]) == "ros2_launch"
+    assert proxy._log_base(["ros2", "launch"]) == "ros2_launch_1"
```
xaru8145: This test the method directly but never exercises the actual dispatch path. It doesn't verify that running the same command twice through `__call__` produces two distinct log files on disk. A bug in `_open_streams` or `__call__` that bypasses `_log_base` entirely would leave this test green while the feature is broken.

src/lambkin/core/shell/proxy.py
```diff
@@ -92,7 +121,89 @@ def __getattr__(self, name: str) -> _CommandProxy:
...
+        the same command has been launched more than once in this iteration.
+
+        Args:
+            argv: The command argv to derive the base name from.
+
+        Returns:
+            A unique base name string, e.g. 'ros2_launch' or 'ros2_launch_1'.
+        """
+        base = "_".join(
+            arg for arg in argv if not arg.startswith("/") and ":=" not in arg
```
xaru8145: What if we pass a path that does not start with "/"?

For example:

```py
ctx.shell.ros2.launch(
    "beluga.launch.xml",
    f"sensor_model:={ctx.variant.sensor_model}",
    f"num_particles:={ctx.variant.num_particles}",
)
```

xaru8145 (changes requested):
PR description mentions `TeeStream` to support `both` log output (console and log output) but I do not see it implemented. I only see either log or console.

The PR description contains a duplicated paragraph: "Manually verified with the Beluga benchmark example across all three output modes..." appears twice. Also, it would be useful to know what manual tests you run and what steps are needed so that I can reproduce it locally.

`Proxy` in dry run mode uses `print`. Please use `logger` as we agreed. Same for beluga benchmark example.

test/core/test_shell.py
teresa-ortega: Solved in 5eb151a

src/lambkin/sdk_options.py
teresa-ortega: Solved in 5eb151a

src/lambkin/core/shell/proxy.py
teresa-ortega: Solved in c8975cd

src/lambkin/core/shell/proxy.py
teresa-ortega: Solved in 49ddfca

test/core/test_shell.py
teresa-ortega: Solved in 7f344de

src/lambkin/core/shell/proxy.py
teresa-ortega: _log_name now derives the log file base name from self._parts (the command tokens) rather than argv, so all positional arguments, whether they are relative paths like beluga.launch.xml, absolute paths, or ROS parameter assignments ,are ignored. The log file will always be named after the command itself, e.g. ros2_launch.stdout.log. Implemented in 8b306df

src/lambkin/core/shell/proxy.py
teresa-ortega: The raise exists to reuse the except subprocess.CalledProcessError block below, which constructs the CommandError with the user-facing message. However, since Popen never raises CalledProcessError on its own, that except block exists solely to catch this manual raise. Another option would be to raise CommandError directly here and remove the except subprocess.CalledProcessError block entirely, which would make the flow more explicit.

src/lambkin/core/process/background.py
```diff
@@ -172,7 +172,7 @@ def background(proxy: CommandProxy, *args: Any, **kwargs: Any) -> BackgroundProc
     with background(ctx.shell.ros2.bag.record, "-O", "output.mcap", "-a"):
         ...
     """
-    argv = proxy.build_argv(*args, **kwargs)
+    argv = proxy._build_argv(*args, **kwargs)
```
xaru8145: Why make this private back again? This was made public in https://github.com/Ekumen-OS/lambkin/pull/151

src/lambkin/core/decorators/benchmark.py
```diff
@@ -111,6 +112,7 @@ def decorator(fn):
 
         @functools.wraps(fn)
         def wrapper(args=None, output_dir=None):
+            logging.basicConfig(level=logging.DEBUG, format="%(message)s")
```
xaru8145: `logging.logging.basicConfig` should be removed from `benchmark.py` as library code should never configure the root logger. 

The right place is at the top of the benchmark function in the user script (`nominal()` in `beluga_benchmark.py`), so logging is configured once at the natural entry point regardless of whether the script is run directly or through the lambkin CLI, and the user has full control over level and format without touching lambkin internals.

src/lambkin/core/shell/proxy.py
```diff
@@ -113,7 +165,74 @@ def get_dry_run(self) -> bool:
...
+        base = self._log_base()
+        out = open(self._cwd / f"{base}.stdout.log", "w")
+        err = open(self._cwd / f"{base}.stderr.log", "w")
+        return out, err
+
+    def _log_name(self) -> str:
+        """Derive a log file base name from the command argv, without any suffix.
+
+        Filters out absolute paths and ROS parameter assignments and joins the
+        remaining tokens with underscores.
```
xaru8145: The docstrings are stale from the previous approach. There are no args for instance.
An example:

```py
   """Derive a log file base name from the command parts.

    Joins the command tokens accumulated so far with underscores.
    Arguments passed at call time are not included, only the tokens
    that form the command itself (e.g. 'ros2_launch').

    Returns:
        A base name string, e.g. 'ros2_launch'.
    """
```

src/lambkin/core/shell/proxy.py
```diff
@@ -113,7 +165,74 @@ def get_dry_run(self) -> bool:
...
+
+        Returns:
+            A tuple of (stdout_file, stderr_file) open for writing.
+        """
+        if self._cwd is None:
+            return None, None
+        base = self._log_base()
+        out = open(self._cwd / f"{base}.stdout.log", "w")
+        err = open(self._cwd / f"{base}.stderr.log", "w")
+        return out, err
```
xaru8145: If ShellProxy is constructed without a cwd and log_output is "file", _open_streams returns (None, None) and _make_popen receives stdout=None, which means inherit from parent — silently behaving as console mode despite the user requesting file mode. This should raise explicitly, for example:
```py
if self._cwd is None:
    raise CommandError(
        self._parts,
        "log_output='file' requires a working directory to be set."
    )
```

xaru8145 (changes requested):
First pass

src/lambkin/core/process/background.py
teresa-ortega: Solved in 554bd95

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in 5b98e1b

src/lambkin/core/shell/proxy.py
teresa-ortega: Solved in fd4880a

src/lambkin/core/shell/proxy.py
teresa-ortega: Solved in 107c198

src/lambkin/core/decorators/benchmark.py
xaru8145: My suggestion is here was not 100% accurate. If we define the logging level in the benchmark, since `nominal()` is called after `cli.py` has already spawned the benchmark script as a subprocess, it will not take any effect.

So let me rephrase my suggestions.

Add --log-level as an SDK option in `sdk_options.py`. Make it very explicit is SDK logging:
```py
    ["--log-level"],
    type=click.Choice(["debug", "info", "warning", "error"], case_sensitive=False),
    default="info",
    help="Log level for lambkin SDK output.",
),
```

Then in `cli.py` parse it before spawning the subprocess and configure logging accordingly. The tricky part is that cli.py uses allow_extra_args=True and passes everything through to the script, so you need to either peek at args for --log-level before forwarding, or handle it as a proper Click option on main() and exclude it from the args forwarded to the script.

This also means nominal() in beluga_benchmark.py no longer needs to call basicConfig itself, the SDK log level controls the CLI process, and the benchmark script process inherits whatever the user configured.

Att. @teresa-ortega

teresa-ortega: You are right that putting basicConfig in nominal() was wrong. However, I think the logging that matters most happens in the benchmark script process, not in cli.py, cli.py does very little, it just spawns the subprocess and waits. So configuring logging in cli.py alone would not give us control over the dry run messages from proxy.py, which run in the script process.
My suggestion would be to do both: parse --log-level in cli.py and configure logging there for the CLI process, and since allow_extra_args=True already forwards all args to the script, benchmark.py can also parse --log-level from the CLI args and call basicConfig with the resolved level before the benchmark loop starts. That way both processes respect the same flag without the user having to configure anything in their script.

xaru8145: Yes, that seems like a good option to me. I agree.

teresa-ortega: Solved in 6fbb9c9

src/lambkin/cli.py
```diff
@@ -99,24 +109,28 @@ def main(script: Path, args: tuple) -> None:
     # TODO(teresa-ortega): Handle concurrent runs, interrupted benchmarks, and re-runs
     # (e.g. detect an already active scope, support partial restarts).
     # To be addressed in phase 6.
-    parent = find_app_slice() or find_delegated_cgroup()
+    parent = find_app_slice()
+    configure_logging(log_level)
+    if parent is None:
+        logger.debug("app.slice not available, falling back to delegated cgroup")
+        parent = find_delegated_cgroup()
```
xaru8145: Please re order as configure_logging is right in between the logic to find the parent cgroup and difficults the reading.

src/lambkin/cli.py
```diff
@@ -90,6 +99,7 @@ def main(script: Path, args: tuple) -> None:
 
     On Ctrl-C, all processes in the benchmark cgroup are terminated and the
     cgroup tree is removed before exiting.
+    logger.debug("app.slice not available at %s", app_slice)
```
xaru8145: Feel like this should not be here

src/lambkin/cli.py
```diff
@@ -99,24 +109,28 @@ def main(script: Path, args: tuple) -> None:
...
-        [sys.executable, str(script), *args],
+        [sys.executable, str(script), "--log-level", log_level, *args],
         start_new_session=True,
         preexec_fn=lambda: (run_cgroup / "cgroup.procs").write_text(str(os.getpid())),
     )
     try:
         proc.wait()
     except KeyboardInterrupt:
-        print("\nInterrupted, cleaning up benchmark processes...", file=sys.stderr)
+        logger.warning("\nInterrupted, cleaning up benchmark processes...")
```
xaru8145: Why the use of "\n" at the beginning?

test/core/test_benchmark.py
```diff
@@ -45,6 +45,8 @@ def fn(ctx):
     assert result == {
         "dry_run": defaults.DRY_RUN,
         "show_options": False,
+        "log_output": "file",
+        "log_level": "info",
```
xaru8145: Use default value for log_level. In all test references.

xaru8145 (changes requested):

examples/beluga/beluga_benchmark.py
```diff
@@ -33,6 +37,8 @@ def nominal(ctx):
     Args:
         ctx: Lambkin context with variant, options, source, and shell access.
     """
+    logging.basicConfig(level=logging.DEBUG, format="%(message)s")
```
xaru8145: Are you sure we want this to be commited?

teresa-ortega: Solved c66b7e6

src/lambkin/cli.py
teresa-ortega: Solved in 43cc858

src/lambkin/cli.py
teresa-ortega: Solved in 43cc858

test/core/test_benchmark.py
teresa-ortega: Solved in 4722e7f

src/lambkin/cli.py
teresa-ortega: Solved in c66b7e6

test/core/test_benchmark.py
```diff
@@ -244,3 +250,23 @@ def fn(ctx):
...
+    """_parse_options returns the correct log_level when provided."""
+
+    def fn(ctx):
+        pass
+
+    result = _parse_options(fn, ["--log-level", "debug"])
+    assert result["log_level"] == "debug"
+
+
+def test_parse_options_log_level_defaults_to_info():
```
xaru8145: Rename the test and documentation as the default could be different from info in the future.

xaru8145 (commented):
Minor thing. 

The rest we iterated over chat regarding exposing lambkin.logger.info and so on...

Once this is addressed, ready to approve.

test/core/test_benchmark.py
teresa-ortega: Solved in 8a015b6

xaru8145 (approved):

PR #155, opened 2026-05-25
Phase 4: Update  Documentation

teresa-ortega:
### Proposed changes

Update documentation to reflect Phase 4 implementation

* Real shell execution
* CLI entrypoint (lambkin command)
* SDK options (--dry-run, --show-options)
* Custom help formatting
* Logging system with per-process output modes
* Background process management via cgroups v2
* Docker and Podman support

#### Type of change

💥 No breaking changes.

### Checklist

### Additional comments
N/A

### DIFF
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -17,8 +17,10 @@ The launch file accepts the map path, sensor model type, and particle count as p
 
 ## Prerequisites
 
-- [Docker](https://docs.docker.com/get-docker/) and
-  [Docker Compose](https://docs.docker.com/compose/)
+- - [Docker](https://docs.docker.com/get-docker/) and
+  [Docker Compose](https://docs.docker.com/compose/), or
+  [Podman](https://podman.io/getting-started/installation) and
+  `podman-compose` (see note below)
 - The following reference files available on the host:
 
 | Artifact | Description |
@@ -28,7 +30,7 @@ The launch file accepts the map path, sensor model type, and particle count as p
 | Groundtruth | Reference trajectory in `.tum` format to evaluate against |
 
 > [!NOTE]
-> `podman-compose` installed via `apt` may be version 1.0.6, which does not support `--podman-run-args`. Install a recent version via `pipx`:
+> `podman-compose` installed via `apt` may be version 1.0.6, which does not support `--profile`. Install a recent version via `pipx`:
 > ```bash
 > sudo apt install pipx
 > pipx ensurepath
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -117,7 +117,7 @@ podman run --rm \
 
 > [!WARNING]
 
-> Both runtimes require elevated privileges to support background process management. Docker runs with `--privileged`, granting the container broad access to host devices and kernel interfaces. Podman uses `--systemd=always`, which allows the container to interact with the host's cgroup v2 hierarchy. LAMBKIN requires these to create transient cgroup scopes that guarantee cleanup of all descendant processes when a benchmark step ends. Only use this in trusted, controlled environments where you own the container invocation — not suitable for shared CI runners or managed cloud environments.
+> Both runtimes require elevated privileges to support background process management. Docker runs with --privileged, granting the container broad access to host devices and kernel interfaces. Podman uses --systemd=always, which allows the container to interact with the host's cgroup v2 hierarchy. LAMBKIN requires these to create transient cgroup scopes that guarantee cleanup of all descendant processes when a benchmark step ends. Only use this in trusted, controlled environments where you own the container invocation — not suitable for shared CI runners or managed cloud environments.
 
 ## Usage
 
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -117,7 +117,7 @@ podman run --rm \
 
 > [!WARNING]
 
-> Both runtimes require elevated privileges to support background process management. Docker runs with *--privileged*', granting the container broad access to host devices and kernel interfaces. Podman uses *--systemd=always*, which allows the container to interact with the host's cgroup v2 hierarchy. LAMBKIN requires these to create transient cgroup scopes that guarantee cleanup of all descendant processes when a benchmark step ends. Only use this in trusted, controlled environments where you own the container invocation — not suitable for shared CI runners or managed cloud environments.
+> Both runtimes require elevated privileges to support background process management. Docker runs with `--privileged`, granting the container broad access to host devices and kernel interfaces. Podman uses `--systemd=always`, which allows the container to interact with the host's cgroup v2 hierarchy. LAMBKIN requires these to create transient cgroup scopes that guarantee cleanup of all descendant processes when a benchmark step ends. Only use this in trusted, controlled environments where you own the container invocation — not suitable for shared CI runners or managed cloud environments.
 
 ## Usage
 
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -78,8 +78,12 @@ docker compose --profile development exec lambkin_dev bash
 **Podman**
 
 ```bash
-podman-compose --profile development up -d
-podman-compose --profile development run --podman-run-args="--systemd=always" --rm lambkin_dev bash
+podman run --rm \
+  --name=lambkin_production \
+  --systemd=always \
+  --network=host \
+  -v "$(pwd)/../results:/ws/examples/beluga/results" \
+  -it lambkin_ros:jazzy bash
 ```
 
 Inside the container:
@@ -100,17 +104,22 @@ Builds a fully self-contained image with all dependencies pre-installed. No manu
 **Docker**
 
 ```bash
+docker compose --profile production up -d
 docker compose --profile production run --rm lambkin_prod bash
 ```
 
 **Podman**
 
 ```bash
-podman-compose --profile production run --podman-run-args="--systemd=always" --rm lambkin_prod bash
+podman run --rm \
+  --name=lambkin_production \
+  --systemd=always \
+  --network=host \
+  -v "$(pwd)/../results:/ws/examples/beluga/results" \
+  -it lambkin_ros:jazzy bash
 ```
 
 > [!WARNING]
-
 > Both runtimes require elevated privileges to support background process management. Docker runs with --privileged, granting the container broad access to host devices and kernel interfaces. Podman uses --systemd=always, which allows the container to interact with the host's cgroup v2 hierarchy.
 
 > LAMBKIN requires these to create transient cgroup scopes that guarantee cleanup of all descendant processes when a benchmark step ends. Only use this in trusted, controlled environments where you own the container invocation; not suitable for shared CI runners or managed cloud environments.
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -117,7 +117,7 @@ podman run --rm \
 
 > [!WARNING]
 
-> Both runtimes require elevated privileges to support background process management. Docker runs with ``--privileged``', granting the container broad access to host devices and kernel interfaces. Podman uses ``--systemd=always``, which allows the container to interact with the host's cgroup v2 hierarchy. LAMBKIN requires these to create transient cgroup scopes that guarantee cleanup of all descendant processes when a benchmark step ends. Only use this in trusted, controlled environments where you own the container invocation — not suitable for shared CI runners or managed cloud environments.
+> Both runtimes require elevated privileges to support background process management. Docker runs with *--privileged*', granting the container broad access to host devices and kernel interfaces. Podman uses *--systemd=always*, which allows the container to interact with the host's cgroup v2 hierarchy. LAMBKIN requires these to create transient cgroup scopes that guarantee cleanup of all descendant processes when a benchmark step ends. Only use this in trusted, controlled environments where you own the container invocation — not suitable for shared CI runners or managed cloud environments.
 
 ## Usage
 
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -17,7 +17,7 @@ The launch file accepts the map path, sensor model type, and particle count as p
 
 ## Prerequisites
 
-- - [Docker](https://docs.docker.com/get-docker/) and
+- [Docker](https://docs.docker.com/get-docker/) and
   [Docker Compose](https://docs.docker.com/compose/), or
   [Podman](https://podman.io/getting-started/installation) and
   `podman-compose` (see note below)
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -122,7 +122,7 @@ podman-compose --profile production run --podman-run-args="--systemd=always" --r
 ```
 
 > [!WARNING]
-> Both runtimes require elevated privileges to support background process management. Docker runs with --privileged, granting the container broad access to host devices and kernel interfaces. Podman uses --systemd=always, which allows the container to interact with the host's cgroup v2 hierarchy. LAMBKIN requires these to create transient cgroup scopes that guarantee cleanup of all descendant processes when a benchmark step ends. Only use this in trusted, controlled environments where you own the container invocation; not suitable for shared CI runners or managed cloud environments.
+> Both runtimes require elevated privileges to support background process management. Docker runs with --privileged, granting the container broad access to host devices and kernel interfaces. Podman uses --systemd=always, which allows the container to interact with the host's cgroup v2 hierarchy.
 
 ## Usage
 
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -44,26 +44,25 @@ volumes:
 
 ### **2. Build the image**
 
-
-####  Development
+#### Development
 
 ```bash
 cd examples/beluga/docker
-docker compose --profile development build  # Docker
-podman compose --profile development build  # Podman
+docker compose --profile development build   # Docker
+podman-compose --profile development build   # Podman
 ```
 
-####  Production
+#### Production
 
 ```bash
 cd examples/beluga/docker
-docker compose --profile production build  # Docker
-podman compose --profile production build  # Podman
+docker compose --profile production build    # Docker
+podman-compose --profile production build    # Podman
 ```
 
 ### **3. Start the container**
 
-Two Docker profiles are available depending on your use case.
+Two profiles are available depending on your use case.
 
 #### Development
 
@@ -79,8 +78,8 @@ docker compose --profile development exec lambkin_dev bash
 **Podman**
 
 ```bash
-podman compose --profile development up -d
-podman compose --profile development exec lambkin_dev bash
+podman-compose --profile development up -d
+podman-compose --profile development run --podman-run-args="--systemd=always" --rm lambkin_dev bash
 ```
 
 Inside the container:
@@ -101,19 +100,14 @@ Builds a fully self-contained image with all dependencies pre-installed. No manu
 **Docker**
 
 ```bash
-docker compose -f docker-compose.yml --profile production run --rm lambkin_prod bash
+docker compose --profile production run --rm lambkin_prod bash
 ```
+
 **Podman**
 
 ```bash
-podman run --rm \
-  --name=lambkin_production \
-  --systemd=always \
-  --network=host \
-  -v "$(pwd)/../results:/ws/examples/beluga/results" \
-  -it lambkin_ros:jazzy bash
+podman-compose --profile production run --podman-run-args="--systemd=always" --rm lambkin_prod bash
 ```
-
 
 > [!WARNING]
 
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -116,7 +116,9 @@ podman run --rm \
 
 
 > [!WARNING]
-Both runtimes require elevated privileges to support background process management. Docker runs with --privileged, granting the container broad access to host devices and kernel interfaces. Podman uses --systemd=always, which allows the container to interact with the host's cgroup v2 hierarchy. LAMBKIN requires these to create transient cgroup scopes that guarantee cleanup of all descendant processes when a benchmark step ends. Only use this in trusted, controlled environments where you own the container invocation — not suitable for shared CI runners or managed cloud environments.
+
+> Both runtimes require elevated privileges to support background process management. Docker runs with ``--privileged``', granting the container broad access to host devices and kernel interfaces. Podman uses ``--systemd=always``, which allows the container to interact with the host's cgroup v2 hierarchy. LAMBKIN requires these to create transient cgroup scopes that guarantee cleanup of all descendant processes when a benchmark step ends. Only use this in trusted, controlled environments where you own the container invocation — not suitable for shared CI runners or managed cloud environments.
+
 ## Usage
 
 Inside the Docker container/enviroment, run the following command:
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -130,5 +130,3 @@ uv run lambkin examples/beluga/beluga_benchmark.py
 ```
 
 Once complete, results are written to results/ organized by configuration (`var_<number>/iter_<number>`). Each iteration contains the recorded bag, TUM trajectory, and APE metrics.
-
-> Note: Full execution is not yet implemented. At this stage the benchmark runs in dry-run mode only, printing all commands that would be executed without running them.
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -78,12 +78,8 @@ docker compose --profile development exec lambkin_dev bash
 **Podman**
 
 ```bash
-podman run --rm \
-  --name=lambkin_production \
-  --systemd=always \
-  --network=host \
-  -v "$(pwd)/../results:/ws/examples/beluga/results" \
-  -it lambkin_ros:jazzy bash
+podman-compose --profile development up -d
+podman-compose --profile development run --podman-run-args="--systemd=always" --rm lambkin_dev bash
 ```
 
 Inside the container:
@@ -111,12 +107,7 @@ docker compose --profile production run --rm lambkin_prod bash
 **Podman**
 
 ```bash
-podman run --rm \
-  --name=lambkin_production \
-  --systemd=always \
-  --network=host \
-  -v "$(pwd)/../results:/ws/examples/beluga/results" \
-  -it lambkin_ros:jazzy bash
+podman-compose --profile production run --podman-run-args="--systemd=always" --rm lambkin_prod bash
 ```
 
 > [!WARNING]
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -120,9 +120,7 @@ podman run --rm \
 ```
 
 > [!WARNING]
-> Both runtimes require elevated privileges to support background process management. Docker runs with --privileged, granting the container broad access to host devices and kernel interfaces. Podman uses --systemd=always, which allows the container to interact with the host's cgroup v2 hierarchy.
-
-> LAMBKIN requires these to create transient cgroup scopes that guarantee cleanup of all descendant processes when a benchmark step ends. Only use this in trusted, controlled environments where you own the container invocation; not suitable for shared CI runners or managed cloud environments.
+> Both runtimes require elevated privileges to support background process management. Docker runs with --privileged, granting the container broad access to host devices and kernel interfaces. Podman uses --systemd=always, which allows the container to interact with the host's cgroup v2 hierarchy. LAMBKIN requires these to create transient cgroup scopes that guarantee cleanup of all descendant processes when a benchmark step ends. Only use this in trusted, controlled environments where you own the container invocation; not suitable for shared CI runners or managed cloud environments.
 
 ## Usage
 
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -27,6 +27,15 @@ The launch file accepts the map path, sensor model type, and particle count as p
 | Map | Static map file in `.yaml` and `.pgm` format |
 | Groundtruth | Reference trajectory in `.tum` format to evaluate against |
 
+> [!NOTE]
+> `podman-compose` installed via `apt` may be version 1.0.6, which does not support `--podman-run-args`. Install a recent version via `pipx`:
+> ```bash
+> sudo apt install pipx
+> pipx ensurepath
+> pipx install podman-compose
+> ```
+> Verify the installed version with `podman-compose --version` before proceeding.
+
 ## Setup
 
 If you are running the benchmark as-is, use the Production profile. If you are modifying the benchmark script or the ROS 2 package, use the Development profile.
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -117,7 +117,9 @@ podman run --rm \
 
 > [!WARNING]
 
-> Both runtimes require elevated privileges to support background process management. Docker runs with --privileged, granting the container broad access to host devices and kernel interfaces. Podman uses --systemd=always, which allows the container to interact with the host's cgroup v2 hierarchy. LAMBKIN requires these to create transient cgroup scopes that guarantee cleanup of all descendant processes when a benchmark step ends. Only use this in trusted, controlled environments where you own the container invocation — not suitable for shared CI runners or managed cloud environments.
+> Both runtimes require elevated privileges to support background process management. Docker runs with --privileged, granting the container broad access to host devices and kernel interfaces. Podman uses --systemd=always, which allows the container to interact with the host's cgroup v2 hierarchy.
+
+> LAMBKIN requires these to create transient cgroup scopes that guarantee cleanup of all descendant processes when a benchmark step ends. Only use this in trusted, controlled environments where you own the container invocation; not suitable for shared CI runners or managed cloud environments.
 
 ## Usage
 
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -126,7 +126,7 @@ podman run --rm \
 Inside the Docker container/enviroment, run the following command:
 
 ```bash
-uv run examples/beluga/beluga_benchmark.py
+uv run lambkin examples/beluga/beluga_benchmark.py
 ```
 
 Once complete, results are written to results/ organized by configuration (`var_<number>/iter_<number>`). Each iteration contains the recorded bag, TUM trajectory, and APE metrics.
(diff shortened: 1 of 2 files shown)


examples/beluga/README.md
```diff
@@ -67,11 +69,20 @@ Two Docker profiles are available depending on your use case.
...
+
 ```bash
 docker compose --profile development up -d
 docker compose --profile development exec lambkin_dev bash
 ```
 
+**Podman**
+
+```bash
+podman compose --profile development up -d
```
xaru8145: AFAIK this is not the way to run the podman compose as it is unable to read the systemd=always flag. From what I remember this should be `podman-compose` using `--podman-run-args="--systemd=always"`

examples/beluga/README.md
```diff
@@ -87,19 +98,35 @@ source install/setup.bash
...
+
 ```bash
-docker compose --profile production up -d
-docker compose --profile production exec lambkin_prod bash
+podman run --rm \
+  --name=lambkin_production \
+  --systemd=always \
+  --network=host \
+  -v "$(pwd)/../results:/ws/examples/beluga/results" \
+  -it lambkin_ros:jazzy bash
```
xaru8145: I would prioritize/document the podman-compose option as it has the examples of the mounted volumes the user can configure once. If the user want to use podman run then it needs to pass the volumes manually every time, which is more tedious.

src/lambkin/README.md
```diff
@@ -44,11 +44,76 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+├── ros2_launch.stderr.log
+├── ros2_bag_record.stdout.log
+└── ros2_bag_record.stderr.log
+```
+
+**Precedence** (highest to lowest)
+
+1. **Per-call** — `_log_output` keyword at the call site
+2. **Benchmark option** — `@lambkin.option("--log-output", default="file")`
+3. **ShellProxy default** — `ShellProxy(log_output="file")`
```
xaru8145: - I think the precedence is not right. Isn't CLI the highest?
- log output is not registered via @lambkin.option as it is an SDK integrated option.
- log_output in per call has a leading slash: `_log_output` should be `log_output`

src/lambkin/README.md
xaru8145:
```suggestion
@lambkin.option("--clock-rate", default=1)
```

xaru8145:
```suggestion
    with lambkin.process.background(ctx.shell.my_recorder, "-o", "output"):
```

src/lambkin/README.md
```diff
@@ -44,11 +44,76 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
 
 Abstracts shell command dispatch. Exposes the host environment's executables as Python attributes — accessing `shell.my_tool` returns a callable that runs `my_tool` with the given arguments, letting benchmark scripts invoke external processes without hardcoding paths or constructing subprocess calls manually. Accessible through the context.
 
+**Background Process**
+
+A context manager that wraps a Shell command and manages its full lifecycle — start, monitor, and clean up — ensuring no orphaned processes survive when the benchmark ends or is interrupted. Uses cgroups v2 to guarantee kernel-level cleanup of the entire process tree, including descendants that have detached via setsid or setpgid. Used via lambkin.process.background(...)
+
+**Process Isolation with cgroups v2**
+LAMBKIN uses cgroups v2 to track and clean up every process spawned during a benchmark run. Unlike process groups or sessions, a process cannot escape its cgroup by calling setsid or setpgid — the kernel enforces containment regardless of what the process does. This makes it the only reliable mechanism for cleaning up a full ROS 2 process tree.
```
xaru8145: I wouldn't mention a "full ROS 2 process tree". Remember LAMBKIN is ROS agnostic. a ROS 2 process tree is just an example but not the only one.

src/lambkin/README.md
```diff
@@ -44,11 +44,76 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
 
 Abstracts shell command dispatch. Exposes the host environment's executables as Python attributes — accessing `shell.my_tool` returns a callable that runs `my_tool` with the given arguments, letting benchmark scripts invoke external processes without hardcoding paths or constructing subprocess calls manually. Accessible through the context.
 
+**Background Process**
+
+A context manager that wraps a Shell command and manages its full lifecycle — start, monitor, and clean up — ensuring no orphaned processes survive when the benchmark ends or is interrupted. Uses cgroups v2 to guarantee kernel-level cleanup of the entire process tree, including descendants that have detached via setsid or setpgid. Used via lambkin.process.background(...)
```
xaru8145:
```suggestion
A context manager that wraps a Shell command and manages its full lifecycle — start, monitor, and clean up — ensuring no orphaned processes survive when the benchmark ends or is interrupted. Uses cgroups v2 to guarantee kernel-level cleanup of the entire process tree, including descendants that have detached via setsid or setpgid. Used via `lambkin.process.background(...)`
```

src/lambkin/README.md
```diff
@@ -44,11 +44,76 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+
+**Process Isolation with cgroups v2**
+LAMBKIN uses cgroups v2 to track and clean up every process spawned during a benchmark run. Unlike process groups or sessions, a process cannot escape its cgroup by calling setsid or setpgid — the kernel enforces containment regardless of what the process does. This makes it the only reliable mechanism for cleaning up a full ROS 2 process tree.
+The cgroup hierarchy for a run looks like this:
+```
+app.slice/                               ← user's systemd app slice
```
xaru8145, line 60: Mention that when no user app slice is found we attempt to find a delegated cgroup. And put an example. Running in the host guarantees a user systemd app slice being available but when running in containerized environments...

src/lambkin/README.md
xaru8145, line 1: I am missing a comment explaining or emphasizing the shell usage difference in background vs used standalone (parenthesis).

src/lambkin/README.md
xaru8145: Split these two sections more cleanly  (Background and Process Isolation with cgroups v2), as they currently overlap and leave gaps.

**BackgroundProcess** should lead with what it actually does for the user: it lets you run a process in background while the benchmark continues executing — that's its primary purpose and it's not stated. Then clarify the cleanup scope: it terminates the process and its descendants when the context exits, but only that subtree. A process spawned inside a background() block that itself forks children is covered; anything else in the iteration is not. The current wording 'ensures no orphaned processes survive' overpromises — if a process is launched outside a background() context manager, this class does nothing for it.

**Process Isolation with cgroups v2** should own two distinct ideas that are currently missing or blurred:
- Isolation: each iteration runs in its own cgroup, meaning processes from one iteration cannot bleed into another. If iteration N is still cleaning up when iteration N+1 starts, there is no interference. This justifies why we create a per-iteration cgroup rather than relying solely on BackgroundProcess.
- Full-run cleanup guarantee: the iteration cgroup covers every process spawned during that run, not just those wrapped in background(). A foreground call still lands in the iteration cgroup. This is the safety net that BackgroundProcess alone cannot provide, and it's the most important thing a user should understand about the cgroup design — even a benchmark that never calls background() gets this guarantee for free.

src/lambkin/README.md
xaru8145, line 232: Missing the metadata file here. Worth also adding log file examples.

xaru8145 (changes requested):

src/lambkin/README.md
xaru8145: Worth highlighting:  the cgroup design gives you process lifetime containment, not communication isolation. So if we run iterations in parallel, there will be cross talk in between nodes.

src/lambkin/README.md
teresa-ortega: Solved in 620ffde

src/lambkin/README.md
teresa-ortega: Solved in 620ffde

src/lambkin/README.md
teresa-ortega: Solved in 620ffde

src/lambkin/README.md
teresa-ortega: Solved in e26f77b

src/lambkin/README.md
teresa-ortega: Solved in 52d4a25

src/lambkin/README.md
teresa-ortega: Solved in a409d7d

examples/beluga/README.md
teresa-ortega: Solved in 037e7d1

examples/beluga/README.md
teresa-ortega: Solved in 037e7d1

src/lambkin/README.md
teresa-ortega: Solved in 105a3e3

src/lambkin/README.md
teresa-ortega: Solved in  105a3e3

teresa-ortega: Solved in 105a3e3

src/lambkin/README.md
xaru8145: I don't think this is is fully addressed. The first comment has not been taken into account and the second one is added as a note that is not properly visualized.

src/lambkin/README.md
```diff
@@ -44,11 +44,111 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+                  not executed.
+  --show-options  List all SDK and custom options available for this benchmark
+                  script and exit.
+
+Custom Options (script-defined):
+  Options registered in your benchmark script via @lambkin.option.
+  Run 'lambkin SCRIPT --show-options' to list them.
+```
+
+## Logging
```
xaru8145, line 110: I am missing the log level control over the sdk logs. Maybe you can separate here into 2 subsections for clarity. One that addressed SDK logging another that addresses Process logging.

src/lambkin/README.md
```diff
@@ -44,11 +44,111 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+```python
+# Correct — proxy passed without calling it
+with lambkin.process.background(
+    ctx.shell.ros2.launch, "beluga_ros2", "beluga.launch.py"
+):
+    ...
+
+# Wrong — calling it with () executes it immediately as a foreground process
+with lambkin.process.background(
+    ctx.shell.ros2.launch(), "beluga_ros2", "beluga.launch.py"
```
xaru8145:
```suggestion
    ctx.shell.ros2.launch("beluga_ros2", "beluga.launch.py")
```

src/lambkin/README.md
```diff
@@ -44,11 +44,111 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
 
 Abstracts shell command dispatch. Exposes the host environment's executables as Python attributes — accessing `shell.my_tool` returns a callable that runs `my_tool` with the given arguments, letting benchmark scripts invoke external processes without hardcoding paths or constructing subprocess calls manually. Accessible through the context.
 
+**Background Process**
+
+Runs a process in the background while the benchmark continues executing. Takes a shell command (without calling it) and manages its full lifecycle — start, monitor, and clean up — as a context manager. When the context exits, it terminates the process and all its descendants; any process spawned outside a `background()` block is not covered.
+
+> [!WARNING]
+> ctx.shell builds commands lazily through attribute chaining — each attribute access appends a word to the command. Pass the proxy without calling it to `lambkin.process.background()`. Calling it with () executes it immediately as a foreground process before `background()` can manage it.
```
xaru8145: Sorry but this does not very read very clearly to me.

```suggestion
> [!WARNING]
> Pass the command proxy to background() without calling it — ctx.shell.my_tool, not ctx.shell.my_tool(). Calling it with () runs the process immediately as a foreground blocking call and background() will raise an error.
```

With this comment probably the example below is not needed.

src/lambkin/README.md
teresa-ortega: Solved in e230fce

src/lambkin/README.md
```diff
@@ -44,11 +44,111 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+
+# Wrong — calling it with () executes it immediately as a foreground process
+with lambkin.process.background(
+    ctx.shell.ros2.launch(), "beluga_ros2", "beluga.launch.py"
+):
+    ...
+```
+
+**Process Isolation with cgroups v2**
+LAMBKIN places each iteration in its own cgroup, so every process spawned during that iteration — whether inside a `background()` block or not — is tracked and contained. This serves two purposes:
```
xaru8145: I don't think this is how you want this to look.

Image

Anyway, I think this diserves a section of its own, as it is not a core concept. Maybe a section named "Cgroups Process Management".

src/lambkin/README.md
```diff
@@ -44,11 +44,111 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+
+* Isolation — processes from one iteration cannot bleed into another. If iteration N is still cleaning up when iteration N+1 starts, there is no interference.
```
xaru8145: This is misleading. The core idea here is that the actual guarantee is narrower: processes from a previous iteration won't survive into the next one because the cgroup is torn down on Context.__exit__. But that is already explained above.

src/lambkin/README.md
```diff
@@ -44,11 +44,111 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+    └── iter-var_1-iter_1-e5f6a7b8/
+        ├── my_algorithm-a9b0c1d2/
+        └── my_recorder-e3f4a5b6/
+```
+
+Cleanup on exit. When a background() context exits normally, LAMBKIN sends `SIGTERM` to all processes in the cgroup, waits for a grace period, then sends `SIGKILL` to any survivors. On Ctrl-C, the CLI writes 1 to `cgroup.kill`, which the kernel propagates instantly to the entire tree.
+
+
+> [!NOTE]
+> The cgroup design provides process lifetime containment, not communication isolation. If iterations were to run in parallel, processes from different iterations could still communicate with each other.
```
xaru8145: Feels more like a warning

src/lambkin/README.md
```diff
@@ -65,16 +165,17 @@ A minimal example composing the SDK primitives described above into a working be
...
         param_a=["x", "y"],
         param_b=[1, 10, 100],
     ),
     num_iterations=5,
 )
-@lambkin.option("--clock-rate", default=100.0)
+@lambkin.option("--clock-rate", default=1.0)
 def my_benchmark(ctx):
-    with lambkin.process.background(ctx.shell.my_recorder, "-O", "output.mcap"):
+    with lambkin.process.background(ctx.shell.my_recorder, "-", "output.mcap"):
```
xaru8145: What happened with -o?
```suggestion
    with lambkin.process.background(ctx.shell.my_recorder, "-o", "output.mcap"):
```

src/lambkin/README.md
```diff
@@ -83,28 +184,45 @@ def my_benchmark(ctx):
...
+
 @my_benchmark.input
 def dataset(ctx):
     return ctx.source.path.parent / "datasets" / "my_dataset.mcap"
 
+
 if __name__ == "__main__":
     my_benchmark()
+```
+**Run it with the CLI:**
```
xaru8145: Be consistent with the style below:
```suggestion
Run it with the CLI:
```

xaru8145 (changes requested):

src/lambkin/README.md
teresa-ortega: Solved in 819b039

src/lambkin/README.md
teresa-ortega: Solved in fe356c4

src/lambkin/README.md
teresa-ortega: Solved in 347267c

src/lambkin/README.md
teresa-ortega: Solved in 5ddea14

src/lambkin/README.md
teresa-ortega: Solved in 060aa7e

src/lambkin/README.md
teresa-ortega: Solved in 09ede79

src/lambkin/README.md
teresa-ortega: This section had been deleted

src/lambkin/README.md
teresa-ortega: Solved in 2a1243a

src/lambkin/README.md
```diff
@@ -44,11 +44,123 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+    └── iter-var_1-iter_1-e5f6a7b8/
+        ├── my_algorithm-a9b0c1d2/
+        └── my_recorder-e3f4a5b6/
+```
+
+Cleanup on exit. When a background() context exits normally, LAMBKIN sends `SIGTERM` to all processes in the cgroup, waits for a grace period, then sends `SIGKILL` to any survivors. On Ctrl-C, the CLI writes 1 to `cgroup.kill`, which the kernel propagates instantly to the entire tree.
+
+
+> [!WARNING]
+> The cgroup design provides process lifetime containment, not communication isolation. If iterations were to run in parallel, processes from different iterations could still communicate with each other.
```
xaru8145:
```suggestion
> The cgroup design provides process lifetime containment, not network isolation. If iterations were to run in parallel, processes from different iterations could still communicate with each other.
```

src/lambkin/README.md
```diff
@@ -44,11 +44,123 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+  Options registered in your benchmark script via @lambkin.option.
+  Run 'lambkin SCRIPT --show-options' to list them.
+```
+
+## Logging
+
+LAMBKIN has two independent logging systems: one for its own internal messages and one for subprocess output.
+
+**SDK Logging**
+Controls the verbosity of LAMBKIN's own internal messages. Set via the --log-level SDK option:
```
xaru8145:
```suggestion
### SDK Logging

Controls the verbosity of LAMBKIN's own internal messages. Set via the `--log-level` SDK option:
```

src/lambkin/README.md
```diff
@@ -44,11 +44,123 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+
+| Level | Behavior |
+|------|-----------|
+| `"debug"` | All internal messages visible, including cgroup paths and command dispatch details |
+| `"info"` | Standard operational messages visible (default) |
+| `"info"` | LAMBKIN output suppressed; only warnings and errors shown |
+
+The LAMBKIN logger is fully isolated from the root logger — user scripts can configure their own logging without any interference.
+
+**Process Logging**
```
xaru8145:
```suggestion
### Process Logging
```

src/lambkin/README.md
```diff
@@ -44,11 +44,123 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+Log files are named after the command and written to the iteration directory:
+```
+results/var_1/iter_1/
+├── my_algorithm.stdout.log
+├── my_algorithm.stderr.log
+├── my_recorder.stdout.log
+└── my_recorder.stderr.log
+```
+
+#### Precedence (highest to lowest)
```
xaru8145: I don't think this needs a section.
```suggestion
Precedence (highest to lowest):
```

src/lambkin/README.md
```diff
@@ -44,11 +44,123 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+Controls the verbosity of LAMBKIN's own internal messages. Set via the --log-level SDK option:
+```bash
+lambkin my_benchmark.py --log-level debug
+```
+
+| Level | Behavior |
+|------|-----------|
+| `"debug"` | All internal messages visible, including cgroup paths and command dispatch details |
+| `"info"` | Standard operational messages visible (default) |
+| `"info"` | LAMBKIN output suppressed; only warnings and errors shown |
```
xaru8145: Info twice? Also, there are more than 3 levels: debug, info, warning, error, critical, ...

xaru8145: I am not sure we need to explain the levels of a library that is not ours and is well documented. I would just mention that the logging log levels that can be passed via cli are the ones logging library supports and maybe add a link to their levels documentation. For example https://docs.python.org/3/library/logging.html#logging-levels

maybe mention if it is case sensitive or not. Like if you can pass --log-level DEBUG or --log-level debug

src/lambkin/README.md
```diff
@@ -44,11 +44,123 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
 Abstracts shell command dispatch. Exposes the host environment's executables as Python attributes — accessing `shell.my_tool` returns a callable that runs `my_tool` with the given arguments, letting benchmark scripts invoke external processes without hardcoding paths or constructing subprocess calls manually. Accessible through the context.
 
+**Background Process**
+
+Runs a process in the background while the benchmark continues executing. Takes a shell command (without calling it) and manages its full lifecycle — start, monitor, and clean up — as a context manager. When the context exits, it terminates the process and all its descendants; any process spawned outside a `background()` block is not covered.
+
+> [!WARNING]
+> Pass the command proxy to background() without calling it — ctx.shell.my_tool, not ctx.shell.my_tool(). Calling it with () runs the process immediately as a foreground blocking call and background() will raise an error.
+
+## Process Isolation with cgroups v2
```
xaru8145:
```suggestion
## Process Management with cgroups v2
```

src/lambkin/README.md
```diff
@@ -44,11 +44,123 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+
+LAMBKIN places each iteration in its own cgroup, so every process spawned during that iteration — whether inside a `background()` block or not — is tracked and contained. This serves two purposes:
+
+* Isolation — each iteration runs in its own cgroup, which is torn down when the iteration context exits. Processes from one iteration cannot survive into the next.
+* Full-run cleanup guarantee — the iteration cgroup covers all processes spawned during the run, including foreground calls. Even a benchmark that never uses `background()` gets this guarantee for free.
```
xaru8145: The "Isolation" bullet is still slightly misleading — "each iteration runs in its own cgroup" is just restating the opening sentence, and "cannot survive into the next" is really a cleanup guarantee, not isolation. The two bullets are still blurring into each other.

src/lambkin/README.md
```diff
@@ -44,11 +44,123 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+When running on the host, a user systemd app slice (app.slice) is always available. In containerized environments no app slice may exist — in that case, LAMBKIN falls back to the nearest delegated cgroup it can find. For example, under Podman with `--systemd=always`:
+```
+user.slice/user-1000.slice/user@1000.service/  ← delegated cgroup root
+└── lambkin-my_benchmark-a1b2c3d4/
+    └── iter-var_1-iter_1-e5f6a7b8/
+        ├── my_algorithm-a9b0c1d2/
+        └── my_recorder-e3f4a5b6/
+```
+
+Cleanup on exit. When a background() context exits normally, LAMBKIN sends `SIGTERM` to all processes in the cgroup, waits for a grace period, then sends `SIGKILL` to any survivors. On Ctrl-C, the CLI writes 1 to `cgroup.kill`, which the kernel propagates instantly to the entire tree.
```
xaru8145: Regarding block
> Cleanup on exit. When a background() context exits normally, LAMBKIN sends `SIGTERM` to all processes in the cgroup, waits for a grace period, then sends `SIGKILL` to any survivors. On Ctrl-C, the CLI writes 1 to `cgroup.kill`, which the kernel propagates instantly to the entire tree.

I would promote to a heading of its own under 

```readme
### Process cleanup

My paragraph ...
```

And focus on the iteration cgroup teardown rather than the background.

examples/beluga/README.md
```diff
@@ -87,19 +108,28 @@ source install/setup.bash
...
+
+**Podman**
+
+```bash
+podman-compose --profile production run --podman-run-args="--systemd=always" --rm lambkin_prod bash
 ```
 
+> [!WARNING]
+> Both runtimes require elevated privileges to support background process management. Docker runs with --privileged, granting the container broad access to host devices and kernel interfaces. Podman uses --systemd=always, which allows the container to interact with the host's cgroup v2 hierarchy. LAMBKIN requires these to create transient cgroup scopes that guarantee cleanup of all descendant processes when a benchmark step ends. Only use this in trusted, controlled environments where you own the container invocation; not suitable for shared CI runners or managed cloud environments.
+
```
xaru8145, line 126: I think the ideas were duplicate and there is no need to explain what privilege does. That is for the user to decide if it is suitable or not for their use.
```suggestion
> [!WARNING]
> Both runtimes require elevated privileges to support background process management. Docker runs with --privileged, granting the container broad access to host devices and kernel interfaces. Podman uses --systemd=always, which allows the container to interact with the host's cgroup v2 hierarchy.
```

xaru8145 (changes requested):

examples/beluga/README.md
teresa-ortega: Solved in ef09ce0

src/lambkin/README.md
teresa-ortega: Solved in bf8384b

src/lambkin/README.md
teresa-ortega: Solved in 7203bcd

src/lambkin/README.md
teresa-ortega: Solved in ab66d53

src/lambkin/README.md
teresa-ortega: Solved in ee780c6

teresa-ortega: Solved in ee780c6

src/lambkin/README.md
teresa-ortega: Solved in 2b2d956

src/lambkin/README.md
teresa-ortega: Solved in ee61fa3

src/lambkin/README.md
teresa-ortega: Solved in d571c66

src/lambkin/README.md
teresa-ortega: Solved in 835020a

src/lambkin/README.md
```diff
@@ -44,11 +44,120 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+> Pass the command proxy to background() without calling it — ctx.shell.my_tool, not ctx.shell.my_tool(). Calling it with () runs the process immediately as a foreground blocking call and background() will raise an error.
+
+## Process Management with cgroups v2
+
+LAMBKIN places each iteration in its own cgroup, so every process spawned during that iteration — whether inside a `background()` block or not — is tracked and contained. This serves two purposes:
+
+* Isolation — each iteration runs in a dedicated cgroup, so processes from different iterations cannot interfere with each other.
+* Full-run cleanup guarantee — the iteration cgroup covers all processes spawned during that run, including foreground calls. Even a benchmark that never uses background() gets this guarantee for free.
+
+Unlike process groups or sessions, a process cannot escape its cgroup by calling `setsid` or `setpgid` — the kernel enforces containment regardless.
```
xaru8145: I think this is tighter and less redundant. And we do not orbit around the concept of isolation which can feel tricky here.
```suggestion
LAMBKIN places each iteration in its own cgroup, so every process spawned during that run — whether inside a background() block or not — is tracked and torn down unconditionally when the iteration ends. This means no leftover processes survive into the next iteration, and no process can escape by calling setsid or setpgid — the kernel enforces containment regardless.
```

src/lambkin/README.md
```diff
@@ -44,11 +44,120 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+```
+When running on the host, a user systemd app slice (app.slice) is always available. In containerized environments no app slice may exist — in that case, LAMBKIN falls back to the nearest delegated cgroup it can find. For example, under Podman with `--systemd=always`:
+```
+user.slice/user-1000.slice/user@1000.service/  ← delegated cgroup root
+└── lambkin-my_benchmark-a1b2c3d4/
+    └── iter-var_1-iter_1-e5f6a7b8/
+        ├── my_algorithm-a9b0c1d2/
+        └── my_recorder-e3f4a5b6/
+```
+## Process Cleanup
```
xaru8145: Add a space before the process cleanup title.

src/lambkin/README.md
```diff
@@ -44,11 +44,120 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+
+When a background() context exits normally, LAMBKIN sends `SIGTERM` to all processes in the cgroup, waits for a grace period, then sends `SIGKILL` to any survivors. On Ctrl-C, the CLI writes 1 to `cgroup.kill`, which the kernel propagates instantly to the entire tree.
+
+
+> [!WARNING]
+> The cgroup design provides process lifetime containment, not network isolation. If iterations were to run in parallel, processes from different iterations could still communicate with each other.
```
xaru8145, line 76: Move this block before process cleanup.

src/lambkin/README.md
```diff
@@ -44,11 +44,120 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+Log files are named after the command and written to the iteration directory:
+```
+results/var_1/iter_1/
+├── my_algorithm.stdout.log
+├── my_algorithm.stderr.log
+├── my_recorder.stdout.log
+└── my_recorder.stderr.log
+```
+
+Precedence (highest to lowest)
```
xaru8145:
```suggestion
Precedence (highest to lowest):
```

xaru8145 (changes requested):
Almost there

src/lambkin/README.md
teresa-ortega: Solved in 075f241

src/lambkin/README.md
teresa-ortega: Solved in f672e1d

src/lambkin/README.md
teresa-ortega: Solved in f672e1d

src/lambkin/README.md
teresa-ortega: Solved in a6a5965

src/lambkin/README.md
```diff
@@ -44,11 +44,115 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+└── lambkin-my_benchmark-a1b2c3d4/
+    └── iter-var_1-iter_1-e5f6a7b8/
+        ├── my_algorithm-a9b0c1d2/
+        └── my_recorder-e3f4a5b6/
+```
+
+> [!WARNING]
+> The cgroup design provides process lifetime containment, not network isolation. If iterations were to run in parallel, processes from different iterations could still communicate with each other.
+
+## Process Cleanup
```
xaru8145: This is part of the process management section
```suggestion
### Process Cleanup
```

src/lambkin/README.md
```diff
@@ -44,11 +44,115 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+
+When a background() context exits normally, LAMBKIN sends `SIGTERM` to all processes in the cgroup, waits for a grace period, then sends `SIGKILL` to any survivors. On Ctrl-C, the CLI writes 1 to `cgroup.kill`, which the kernel propagates instantly to the entire tree.
```
xaru8145: Here we are not interested on how background exits but how the entire benchmark is cleaned up per run. Rephrase so we focus on that and not the background.

src/lambkin/README.md
```diff
@@ -44,11 +44,115 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
 
 Abstracts shell command dispatch. Exposes the host environment's executables as Python attributes — accessing `shell.my_tool` returns a callable that runs `my_tool` with the given arguments, letting benchmark scripts invoke external processes without hardcoding paths or constructing subprocess calls manually. Accessible through the context.
 
+**Background Process**
+
+Runs a process in the background while the benchmark continues executing. Takes a shell command (without calling it) and manages its full lifecycle — start, monitor, and clean up — as a context manager. When the context exits, it terminates the process and all its descendants; any process spawned outside a `background()` block is not covered.
+
+> [!WARNING]
+> Pass the command proxy to background() without calling it — ctx.shell.my_tool, not ctx.shell.my_tool(). Calling it with () runs the process immediately as a foreground blocking call and background() will raise an error.
```
xaru8145:
```suggestion
> Pass the command proxy to background() without calling it — `ctx.shell.my_tool`, not `ctx.shell.my_tool()`. Calling it with () runs the process immediately as a foreground blocking call and background() will raise an error.
```

src/lambkin/README.md
```diff
@@ -44,11 +44,115 @@ Registers a CLI option on a benchmark. Built on top of [`click`](https://click.p
...
+
+## Process Management with cgroups v2
+
+LAMBKIN places each iteration in its own cgroup, so every process spawned during that run — whether inside a background() block or not — is tracked and torn down unconditionally when the iteration ends. This means no leftover processes survive into the next iteration, and no process can escape by calling setsid or setpgid — the kernel enforces containment regardless.
```
xaru8145:
```suggestion
LAMBKIN places each iteration in its own cgroup, so every process spawned during that run — whether inside a background() block or not — is tracked and torn down unconditionally when the iteration ends. This means no leftover processes survive into the next iteration, and no process can escape by calling `setsid` or `setpgid` — the kernel enforces containment regardless.
```

xaru8145 (changes requested):

src/lambkin/README.md
teresa-ortega: Solved in 187391e

src/lambkin/README.md
teresa-ortega: Solved in 187391e

src/lambkin/README.md
teresa-ortega: Solved in 187391e

src/lambkin/README.md
xaru8145: Missing this  @teresa-ortega

teresa-ortega: Solvd in 45265bc

xaru8145 (approved):
LGTM

PR #160, opened 2026-06-04
Phase 5: Add variants file and useful logs

xaru8145:
### Proposed changes

This PR adds three log messages to the benchmark execution loop:

- `INFO` before the loop starts with the total number of runs.
- `INFO` per run with the current run count (e.g. 5/40).
- `INFO` per run with the time it took to finish the benchmark.

For the last one, a fully tested function to format the elapsed time into a human user friendly format has been added.

Additional changes focus on writing a `variants.yaml` file to the results base directory mapping each `var_N` folder to its variant parameters, making it easier to identify what each variant folder corresponds to without digging into individual iteration metadata files. 

Documentation is updated accordingly to explain this newly added output. And spacing is standarized.

Minor: adds iteration to the benchmark log info statement.

#### How to Test

Run the benchmark in dry run mode:

```bash
uv run lambkin examples/beluga/beluga_benchmark.py --dry-run
```

Expect to see the 2 newly added info messages and the `results/variants.yaml` file.

#### Type of change

💥 No breaking changes.

### Checklist

### Additional comments

N/A

### DIFF
--- a/examples/beluga/beluga_benchmark.py
+++ b/examples/beluga/beluga_benchmark.py
@@ -33,7 +33,10 @@ def nominal(ctx):
     Args:
         ctx: Lambkin context with variant, options, source, and shell access.
     """
-    lambkin.logger.info("Running benchmark with variant: %s", ctx.variant)
+    lambkin.logger.info(
+        "Running benchmark with variant: %s, iteration: %d", ctx.variant, ctx.iteration
+    )
+
     with lambkin.process.background(
         ctx.shell.ros2.bag.record,
         "--output",
--- a/src/lambkin/core/decorators/benchmark.py
+++ b/src/lambkin/core/decorators/benchmark.py
@@ -31,6 +31,7 @@ import functools
 import inspect
 import logging
 import sys
+import time
 
 import click
 import yaml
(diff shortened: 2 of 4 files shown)


src/lambkin/core/decorators/benchmark.py
```diff
@@ -114,6 +118,9 @@ def wrapper(args=None, output_dir=None):
                 sys.exit(0)
             log_level = options.get("log_level", defaults.LOG_LEVEL)
             configure_logging(log_level)
+            # Calculate total runs for logging purposes.
+            total_runs = len(variants) * num_iterations
+            logger.info("Starting benchmark: %d run(s) total.", total_runs)
             source = Source(path=inspect.getfile(fn))
             # The base context creates a directory for variant 1 / iteration 1
```
teresa-ortega: I think we could mention variants.yaml here, since it's written inside this base_ctx block, something like "used to resolve inputs and write variants.yaml".

examples/beluga/beluga_benchmark.py
```diff
@@ -33,7 +33,10 @@ def nominal(ctx):
     Args:
         ctx: Lambkin context with variant, options, source, and shell access.
     """
-    lambkin.logger.info("Running benchmark with variant: %s", ctx.variant)
+    lambkin.logger.info(
+        "Running benchmark with variant: %s, iteration: %d", ctx.variant, ctx.iteration
```
teresa-ortega, line 37: Consider using `vars(ctx.variant)` instead of `ctx.variant` directly, it avoids the namespace(...) wrapper in the log output, which is a bit noisy.

teresa-ortega (commented):
Could we add a test for variants.yaml? None of the existing tests cover it and it's new functionality. Something like checking that the file exists and that its content matches the expected var_N mapping.

src/lambkin/core/decorators/benchmark.py
xaru8145: I am not sure I understand your comment. Can you expand? You mean add a code comment, a log.info?

examples/beluga/beluga_benchmark.py
xaru8145: The thing is that I do not want the user to know business logic of elements of the context. I think is better that they just print them.

I agree with you that the logging is a not ideal due to the namespace. Another option is overwrite the Namespace __repr__ so that it prints as we want. But that requires creating a thin subclass over SimpleNamespace. What do you think?

src/lambkin/core/decorators/benchmark.py
teresa-ortega: I meant updating the existing base ctx comment to mention that variants.yaml is written inside this block

examples/beluga/beluga_benchmark.py
teresa-ortega: Agreed, the user shouldn't need to know the internal representation. I think creating a subclass of `SimpleNamespace` might add quite a bit of complexity for now. Could be worth keeping a note to revisit this in the future

src/lambkin/core/decorators/benchmark.py
xaru8145: Done in 00caa4ef77fdfc614ca9f15184b9fde8b489c9a9

examples/beluga/beluga_benchmark.py
xaru8145: We can discuss or address in the future. I wouldn't worry much about this as it is just a nice to have.

src/lambkin/core/decorators/benchmark.py
```diff
@@ -41,6 +44,41 @@
...
+    - Under 60 s:  ``'0.0023s'``
+    - Under 1 h:   ``'45m 03s'``
+    - Under 1 day: ``'2h 15m 07s'``
+    - 1 day or more: ``'2d 03h 15m 07s'``
+
+    Sub-second precision is kept only for runs under 60 seconds; longer
+    durations are truncated to whole seconds.
+
+    Args:
+        seconds: Elapsed time in seconds, as returned by ``time.monotonic()``.
```
teresa-ortega:
```suggestion
        seconds (float): Elapsed time in seconds, as returned by ``time.monotonic()``.
```

teresa-ortega (commented):
Minor thing

teresa-ortega (approved):
LGTM

PR #161, opened 2026-06-05
Phase 5: Replace OutputInfo in favor of RunPaths and update defaults

xaru8145:
### Proposed changes

This PR does the following:
- Documents `defaults.py` and makes them type-safe thanks to `Final` qualifier.
- Moves `BENCHMARKS_DIRNAME` constant from context `OutputInfo` into `defaults.py`.
- Replaces `OutputInfo` for `RunPaths` (which is moved to its own file `paths.py`) to avoid collision with incoming output decorator. Init now exports this class.
- Removed TODO comment in Context as now ownership of the path construction logic is moved to `RunPaths` thanks to `from_indices` function. `ctx.output` attribute has been renamed to `ctx.paths`, and `output_dir` constructor attribute to `base_dir`.
- Added log level `critical` choice in `sdk_options.py` as it was missing.
- Tests have been updated accordingly to account for all the changes.

#### Type of change

💥 **Breaking change!** N/A

### Checklist

### Additional comments

N/A

### DIFF
--- a/src/lambkin/core/ctx/context.py
+++ b/src/lambkin/core/ctx/context.py
@@ -38,43 +38,8 @@ from lambkin.core.process.cgroup import (
 )
 from lambkin.core.shell import ShellProxy
 
+from .paths import RunPaths
 from .source import Source
-
-
-class OutputInfo:
-    """Output paths for this variant + iteration.
-
-    Folders are created lazily — only when the path is first accessed.
-
-    Attributes:
-        base_dir: Base output folder for the benchmark (e.g. results/).
-        variant_dir: Root folder for this variant (e.g. results/var_1/).
-        iteration_dir: Folder for the current iteration (e.g. results/var_1/iter_1/).
-    """
-
-    def __init__(self, base_dir: Path, variant_dir: Path, iteration_dir: Path) -> None:
-        """Initialize the output paths for one (variant, iteration) pair.
-
-        Args:
-            base_dir (Path): Base output folder for benchmark.(e.g. results/).
-            variant_dir (Path): Root output folder for this variant
-                (e.g. results/var_1/).
-            iteration_dir (Path): Output folder for the
-                current iteration (e.g. results/var_1/iter_1/).
-        """
-        self.variant_dir = Path(variant_dir)
-        self.base_dir = Path(base_dir)
-        self.iteration_dir = Path(iteration_dir)
-
-
-def _variant_folder_name(index: int) -> str:
-    """Build the variant folder name from its index."""
-    return f"var_{index + 1}"
-
-
-def _iteration_folder_name(iteration: int) -> str:
-    """Build the per-iteration folder name."""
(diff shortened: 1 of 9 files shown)


src/lambkin/core/ctx/context.py
```diff
@@ -121,19 +84,18 @@ def __init__(
...
                 Controls the ``iter_<N>`` subfolder name under the variant
                 directory, where ``N = iteration + 1``.
+            options (dict): Runtime options, as defined by the user.
+                All key-value pairs are exposed as attributes on ``ctx.options``.
             source (Source): Source object describing the benchmark script being
                 executed. Its parent directory is used as the default output
                 directory.
-            options (dict): Runtime options, as defined by the user.
-                All key-value pairs are exposed as attributes on ``ctx.options``.
+            base_dir: Root directory for all benchmark results.
```
teresa-ortega:
```suggestion
            base_dir (Path | str) : Root directory for all benchmark results.
```

src/lambkin/core/ctx/context.py
```diff
@@ -86,27 +51,25 @@ class Context:
...
 
     Attributes:
-        BENCHMARKS_DIRNAME (str): Name for the benchmarks directory.
         variation: Namespaced algorithm parameters for this run.
             All key-value pairs from the variant dict are exposed as attributes.
         source: Source object describing the benchmark script being executed.
         inputs: Namespaced input information.
         options: Namespaced runtime options.
             All key-value pairs from the options dict are exposed as attributes.
         output: Namespaced output paths.
```
teresa-ortega: `output` in the class docstring looks like a leftover from the rename. I think it can be removed since the attribute is now `paths`.

src/lambkin/core/decorators/benchmark.py
teresa-ortega: should `output_dir` in wrapper be renamed to `base_dir` to match the rename, and to be consistent.

```suggestion
        def wrapper(args=None, base_dir=None):
```

teresa-ortega (commented):
Minor things

src/lambkin/core/ctx/context.py
xaru8145: Good catch! Updated in 2f32dabbb59b323c4f11d2b1efe2d506dacdbf65

src/lambkin/core/decorators/benchmark.py
xaru8145: Yes, thanks for the suggestion. Renamed for consistency in 2f32dabbb59b323c4f11d2b1efe2d506dacdbf65

src/lambkin/core/ctx/context.py
xaru8145: Reviewed and updated the class documentation in 2f32dabbb59b323c4f11d2b1efe2d506dacdbf65, as some other attributes were missing!

teresa-ortega (approved):
LGTM

PR #162, opened 2026-06-05
Phase 5: Interrupt foreground process when background dies unexpectedly

teresa-ortega:
### Proposed changes

* Background processes dying while a foreground process blocks the main thread no longer go undetected until the foreground finishes naturally.
* `BackgroundProcess` monitor thread now sets a pending flag in `lambkin.common.signals` and sends `SIGUSR1` to the main thread when it detects an unexpected death.
* `lambkin.common.signals.setup()` registers a `SIGUSR1` handler that raises `LambkinSIGUSR1Interrupt` (a new `BaseException` subclass in `lambkin.common.exceptions`) to unblock `proc.wait()` in `CommandProxy.__call__`. Unrelated `SIGUSR1s` are forwarded to the previous handler.
* `CommandProxy.__call__ `catches `LambkinSIGUSR1Interrupt`, kills the foreground process in the `finally` block, and skips the returncode check, allowing `BackgroundProcess.__exit__` to raise `LambkinProcessDiedUnexpectedlyError` with the correct argv and returncode.
* `benchmark` wrapper calls `signals.setup()` once before the iteration loop.
* Adds integration test `test_background_dies_during_foreground.py` covering the previously untested case.
* Updates `test_exception` and `test_unexpected_death` to use a subprocess foreground instead of `time.sleep` so `SIGUSR1` is handled correctly
* Adds unit test verifying `sigusr1_pending` is set when a background process dies unexpectedly.
* Refactors all integration tests to use `tmp_path` fixture instead of manual `tempfile/shutil` cleanup.

### How to test

**Manual testing**
Add .values() to the beluga launch file to make it fail:
```python

    return LaunchDescription(
        [
            config_file_arg,
            map_path_arg,
            laser_model_arg,
            max_particles_arg,
            beluga_node,
            map_server_node,
            lifecycle_manager_node,
        ].values() # add .values() to make it fail
    ) 
```
Run the beluga_benchmark script 
```bash
```
**Integration test**
```bash
uv run pytest test/integration/test_background_dies_during_foreground.py
```

#### Type of change

💥 No breaking changes

### Checklist

### Additional comments

N/A


src/lambkin/core/decorators/benchmark.py
xaru8145, line 1: Why is the signal handled in the benchmark rather than in cli? Like we do for the keyboard interrupt.

src/lambkin/core/decorators/benchmark.py
```diff
@@ -80,6 +81,19 @@ def _format_elapsed(seconds: float) -> str:
     return f"{m}m {s:02d}s"
 
 
+def _make_sigusr1_handler(previous):
+    def _handle_sigusr1(signum, frame):
+        # SIGUSR1 is a general-purpose signal and could be sent by other
+        # processes or libraries. We only act on it if lambkin set the pending flag.
+        if signals.sigusr1_pending.is_set():
+            signals.sigusr1_pending.clear()
+            raise exceptions.LambkinProcessDiedUnexpectedlyError([], 1)
```
xaru8145: Why do we raise here? The job of the handler is just to manage the signal to interrupt the blocking `proc.wait()` in `CommandProxy.__call__`. Once that's interrupted, the `finally` block kills the foreground process, the `with` block exits, `BackgroundProcess.__exit__` runs, checks `_died_unexpectedly`, and raises `LambkinProcessDiedUnexpectedlyError` with the correct argv and returncode.

xaru8145 (changes requested):

src/lambkin/core/decorators/benchmark.py
teresa-ortega: The CLI and the benchmark script run in separate processes, the CLI launches the script via subprocess.Popen. BackgroundProcess and its monitor thread live in the script's process, so SIGUSR1 is sent there. A handler registered in the CLI would never receive it.

src/lambkin/core/decorators/benchmark.py
teresa-ortega: You are right, the handler should not raise LambkinProcessDiedUnexpectedlyError directly. But proc.wait() is a blocking syscall, the only way to interrupt it from the main thread is by raising an exception from the signal handler.

One option is to introduce an internal exception like _LambkinSIGUSR1Interrupt(BaseException) whose only job is to interrupt proc.wait(). CommandProxy.__call__ catches it in the finally block and swallows it, and BackgroundProcess.__exit__ then raises LambkinProcessDiedUnexpectedlyError with the correct argv and returncode as it already does.

What do you think?

xaru8145: Good approach. One suggestion: name it `LambkinSIGUSR1Interrupt` (no leading underscore) and place it in `lambkin.common.exceptions` alongside `LambkinProcessDiedUnexpectedlyError`. Make sure via test the original exception messages makes it all the way to the terminal.

test/integration/integration_test_background_dies_during_foreground.py
xaru8145, line 1: Update the test so it is automatically picked up by pytest. See https://github.com/Ekumen-OS/lambkin/pull/164 to check what's needed to do it.

src/lambkin/core/decorators/benchmark.py
xaru8145: You're right, I was confused about the process boundary. The handler belongs in the benchmark script process, not the CLI. However, registering it inside `benchmark.py`'s `wrapper()` means it only works when `@benchmark` is used. If someone uses `BackgroundProcess` directly without the decorator, like in the integration tests, they have to duplicate the setup manually each time. Consider extracting this into a dedicated `setup()` function in `lambkin.common.signals` that `wrapper()` calls explicitly:

An illustrative example:

```py
# benchmark.py
from lambkin.common import signals

    signals.setup()  # registers SIGUSR1 handler for the benchmark script process
    ...
```

```py
# lambkin/common/signals.py

sigusr1_pending = threading.Event()

def _make_handler(previous):
    def _handle_sigusr1(signum, frame):
        ...

def setup():
    previous = signal.signal(signal.SIGUSR1, signal.SIG_DFL)
    signal.signal(signal.SIGUSR1, _make_handler(previous))
```

Everything signal-related lives in one place — the pending flag, the handler factory, and the setup function. Make sure you document everything properly in `lambkin/common/signals.py` docstrings and add code comments where appropriate, for example explaining why we need to setup the sigur signal in the benchmark rather than the CLI.

teresa-ortega: Solved in 1b44a4b

test/integration/integration_test_background_dies_during_foreground.py
teresa-ortega: Solved in 1b44a4b

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in 92b621a

src/lambkin/common/signals.py
xaru8145, line 1: Missing the registration of `setup()` function in the `common/__init__`

src/lambkin/common/signals.py
```diff
@@ -0,0 +1,67 @@
...
+        if sigusr1_pending.is_set():
+            sigusr1_pending.clear()
+            raise LambkinSIGUSR1Interrupt
+        elif callable(previous):
+            previous(signum, frame)
+
+    return _handle_sigusr1
+
+
+def setup():
```
xaru8145, line 61: Nothing prevents `setup()` from being called twice, which would chain the handler onto itself. Add a comment in the docstring like "Calling it multiple times will chain handlers, which is harmless but unnecessary."

src/lambkin/common/signals.py
```diff
@@ -0,0 +1,67 @@
...
+            signum: The signal number received.
+            frame: The current stack frame.
+
+        Raises:
+            LambkinSIGUSR1Interrupt: If lambkin set the pending flag before
+                sending SIGUSR1.
+        """
+        if sigusr1_pending.is_set():
+            sigusr1_pending.clear()
+            raise LambkinSIGUSR1Interrupt
```
xaru8145: Instantiating explicitly is more idiomatic and consistent. Both work in Python since raise ExceptionClass is equivalent to raise ExceptionClass().
```suggestion
            raise LambkinSIGUSR1Interrupt()
```

src/lambkin/common/signals.py
```diff
@@ -0,0 +1,67 @@
...
+        elif callable(previous):
+            previous(signum, frame)
```
xaru8145: What happens if `previous` is not callable, like `signal.SIG_DFL`? The `callable(previous)` check silently drops the signal in that case. Pick a consistent policy for unrelated SIGUSR1s and apply it to both cases — either forward faithfully (restore/call whatever was there before) or ignore and log. Right now the behavior is inconsistent: callable previous handlers are forwarded, but SIG_DFL is silently swallowed.

test/integration/test_exception.py
```diff
@@ -30,6 +30,8 @@
 )
 from lambkin.core.shell.proxy import ShellProxy
 
+signals.setup()
```
xaru8145: Module-level code runs at import time, before pytest sets up fixtures or test isolation. `signals.setup()` should be called inside each test function.

test/integration/test_unexpected_death.py
```diff
@@ -30,6 +30,8 @@
 )
 from lambkin.core.shell.proxy import ShellProxy
 
+signals.setup()
```
xaru8145: Module-level code runs at import time, before pytest sets up fixtures or test isolation. `signals.setup()` should be called inside each test function.

test/core/test_background.py
```diff
@@ -136,6 +137,11 @@ def test_background_process_raises_if_dies_unexpectedly(tmp_path):
     iteration_dir.mkdir(parents=True)
     cgroup = make_iteration_cgroup(find_delegated_cgroup(), iteration_dir)
 
+    signal.signal(
+        signal.SIGUSR1,
+        lambda s, f: (_ for _ in ()).throw(LambkinProcessDiedUnexpectedlyError([], 1)),
+    )
```
xaru8145: Why not use `signals.setup()`?

This is a workaround to raise an exception from a lambda. Now that `signals.setup()` exists, this test should just call it instead. It makes it easier to read.

test/integration/test_background_dies_during_foreground.py
```diff
@@ -0,0 +1,80 @@
...
+    signals.setup()
+    iteration_dir = Path(tempfile.mkdtemp())
+    try:
+        cgroup = make_iteration_cgroup(find_delegated_cgroup(), iteration_dir)
+        shell = ShellProxy(dry_run=False, cwd=iteration_dir, cgroup=cgroup)
+        with pytest.raises(LambkinProcessDiedUnexpectedlyError) as exc_info:
+            with background(shell.python3, "-c", DIES_QUICKLY) as _:
+                shell.python3("-c", COOPERATIVE)
+        assert "python3" in str(exc_info.value)
+        assert "1" in str(exc_info.value)
```
xaru8145: Asserting "1" is in the string is weak. It could pass even if the returncode wasn't in the message. Assert on the actual attributes instead:

```suggestion
        assert exc_info.value.argv == ["python3", "-c", DIES_QUICKLY]
        assert exc_info.value.returncode == 1
```

test/integration/test_background_dies_during_foreground.py
xaru8145, line 1: All integration tests use `tempfile.mkdtemp()` with manual `shutil.rmtree` cleanup in finally. Since these are now proper pytest tests, they should use the `tmp_path` fixture which handles cleanup automatically. This is my fault as I introduced this pattern, did not realize till now. Can I ask you to amend for all integration tests?

test/integration/test_unexpected_death.py
xaru8145: I just realized but `DIES_QUICKLY` exits with return code 0, not a non-zero code. `LambkinProcessDiedUnexpectedlyError` is raised based on unexpected death regardless of return code, so the test passes. But it's misleading as "dies unexpectedly" implies a crash, not a clean exit. Should use `sys.exit(1)` like in the other test files to make the intent clear

test/core/test_background.py
xaru8145: If the signal is requires for many tests across the test suite, consider using a fixture. If all tests needed, you can use `autouse=True`.

test/core/test_background.py
xaru8145, line 1: The unit tests don't verify that `signals.sigusr1_pending` is set when a background process dies unexpectedly. This is a key part of the SIGUSR1 mechanism and should be tested directly in here in `test_background.py`.

xaru8145 (changes requested):

src/lambkin/common/signals.py
teresa-ortega: Solved in c1b8605

src/lambkin/common/signals.py
teresa-ortega: Solved in c1b8605

src/lambkin/common/signals.py
teresa-ortega: Solved in 70283af

src/lambkin/common/signals.py
teresa-ortega: Solved in 7207a39

test/integration/test_exception.py
teresa-ortega: Solved in 75a5dc1

test/integration/test_unexpected_death.py
teresa-ortega: Solved in 75a5dc1

test/core/test_background.py
teresa-ortega: Solved in 8856324

test/integration/test_background_dies_during_foreground.py
teresa-ortega: Solved in 3062f3a

test/integration/test_background_dies_during_foreground.py
teresa-ortega: Solved in efb1121

test/integration/test_unexpected_death.py
teresa-ortega: Solved in efb1121

test/core/test_background.py
teresa-ortega: Solved in efb1121

src/lambkin/common/__init__.py
```diff
@@ -17,12 +17,15 @@
 Provides defaults, named products, and base exceptions.
 """
 
-from lambkin.common import defaults, exceptions
+from lambkin.common import defaults, exceptions, signals
+from lambkin.common.signals import setup
```
xaru8145: I was wrong about my previous suggestion of exposing setup. 

Importing setup directly into lambkin.common's namespace via from lambkin.common.signals import setup causes signals.py to be executed at import time as a side effect of importing lambkin.common. This means sigusr1_pending = threading.Event() is created in every module that imports lambkin.common, even those that have nothing to do with signals.

Right now that's harmless, but if signals.py ever gains more module-level initialization it would silently affect the entire codebase. Drop the explicit re-export and let callers import signals explicitly when they need it.

Sorry for the confusion.

src/lambkin/common/__init__.py
```diff
@@ -17,12 +17,15 @@
 Provides defaults, named products, and base exceptions.
```
xaru8145: Update the docu to reference signals.

test/core/test_background.py
```diff
@@ -204,3 +202,30 @@ def test_background_console_mode_creates_no_log_files(tmp_path):
     with background(shell.sleep, "30", log_output="console"):
         pass
     assert not list(iteration_dir.glob("*.log"))
+
+
+def test_background_process_sets_sigusr1_pending_on_unexpected_death(tmp_path):
+    """sigusr1_pending is set by the monitor thread when a process dies unexpectedly."""
+    from lambkin.common import signals as lambkin_signals
```
xaru8145: This is not needed as it is already imported in the top of the file.

test/core/test_background.py
```diff
@@ -204,3 +202,30 @@ def test_background_console_mode_creates_no_log_files(tmp_path):
     with background(shell.sleep, "30", log_output="console"):
         pass
     assert not list(iteration_dir.glob("*.log"))
+
+
+def test_background_process_sets_sigusr1_pending_on_unexpected_death(tmp_path):
```
xaru8145, line 213: The signal handler is not restored after the test, leaving _capturing_handler installed for all subsequent tests in the module. Save and restore the previous handler to guarantee test isolation regardless of whether the test passes or fails.

xaru8145 (changes requested):
Almost there

src/lambkin/common/__init__.py
teresa-ortega: Don't worry, solved in 1d70344

src/lambkin/common/__init__.py
teresa-ortega: Solved in 1d70344

test/core/test_background.py
teresa-ortega: Solved in 5c0144f

test/core/test_background.py
teresa-ortega: Solved in 5c0144f

test/core/test_background.py
```diff
@@ -198,3 +208,30 @@ def test_background_console_mode_creates_no_log_files(tmp_path):
...
+    try:
+        shell = ShellProxy(dry_run=False, cwd=tmp_path, cgroup=cgroup)
+        with pytest.raises(LambkinProcessDiedUnexpectedlyError):
+            with background(shell.sleep, "0"):
+                shell.python3("-c", COOPERATIVE)
+        assert pending_was_set and pending_was_set[0], (
+            "sigusr1_pending should be set before the signal handler runs"
+        )
+    finally:
+        signal.signal(signal.SIGUSR1, previous)
```
xaru8145, line 237: Missing the clear of `sigusr1_pending`.
```suggestion
        signal.signal(signal.SIGUSR1, previous)
        signals.sigusr1_pending.clear()
```

xaru8145 (changes requested):
Minor thing

test/core/test_background.py
teresa-ortega: Solved in 78cf4da

xaru8145 (approved):
LGTM

PR #166, opened 2026-06-09
Phase 5: Add support to partial restarts via cache

xaru8145:
### Proposed changes

Implements a Docker-inspired partial restart cache for the benchmark execution loop. Each `(variant, iteration)` pair is identified by a stable hash of its inputs (variant parameters, iteration index, and user-defined options). Completion is persisted to `lambkin_metadata.yaml` on disk, so successfully completed iterations are automatically skipped on subsequent runs without any manual intervention.

This addresses a key pain point in long benchmarks: if a run is interrupted or a single iteration fails, the entire sweep no longer needs to restart from scratch.

Key implementation details:

- `lambkin/core/ctx/cache.py` introduces `compute_run_hash` and `is_completed`, keeping cache logic cleanly separated for the context although it is related.
- `Context.__init__` is now lightweight — all side effects (directory creation, cgroup setup, metadata write) are deferred to `__enter__`, which checks the cache first and skips setup entirely on a hit.
- `_setup_directories` removes and recreates `iteration_dir` if it exists, cleaning up leftover artifacts from a previous failed run such as a partial `ros2 bag record` output folder.
- The base context used to resolve inputs is constructed but never entered, so it no longer creates `var_1/iter_1` as a side effect before the loop.
- `--no-cache flag` added to SDK options as an escape hatch to force a full rerun.
- Source file contents are intentionally excluded from the hash to avoid invalidating the cache on cosmetic changes such as added log statements or comments.

#### How to test

Run the full benchmark once to completion, then run it again and verify that all iterations are skipped:

```bash
lambkin my_benchmark.py
lambkin my_benchmark.py  # should log cache hits for all iterations
```

To force a full rerun:

```bash
lambkin my_benchmark.py --no-cache
```

To simulate a partial restart, interrupt a running benchmark with Ctrl-C and rerun — only incomplete iterations should execute.

#### Type of change

💥 **Breaking change!** 

`Context` no longer creates directories, cgroups, or metadata on construction. All side effects are deferred to `__enter__`. Any code constructing Context outside a `with` block and accessing `ctx.shell` or filesystem state will break. Always use `with Context(...) as ctx:`.

### Checklist

### Additional comments

N/A

### DIFF
--- a/src/lambkin/core/ctx/context.py
+++ b/src/lambkin/core/ctx/context.py
@@ -31,11 +31,7 @@ from typing import Any
 
 import yaml
 
-from lambkin.common import (
-    compute_run_hash,
-    defaults,
-    is_completed,
-)
+from lambkin.common import defaults
 from lambkin.core.process.cgroup import (
     find_delegated_cgroup,
     kill_cgroup_tree,
@@ -44,6 +40,7 @@ from lambkin.core.process.cgroup import (
 )
 from lambkin.core.shell import ShellProxy
 
+from .cache import compute_run_hash, is_completed
 from .paths import RunPaths
 from .source import Source
 
--- a/src/lambkin/core/ctx/context.py
+++ b/src/lambkin/core/ctx/context.py
@@ -67,6 +67,11 @@ class Context:
         paths: Output paths for this (variant, iteration) run.
         shell: ShellProxy instance configured for this run, with working directory
             set to the iteration output folder and cgroup set to the iteration cgroup.
+            Only available after ``__enter__`` is called on a cache miss. Accessing
+            it on a cache hit (``ctx.skipped is True``) raises ``AttributeError``.
+        skipped: True if this iteration was skipped due to a cache hit. Always
+            check ``ctx.skipped`` before accessing ``ctx.shell`` or writing any
+            outputs.
     """
 
     METADATA_FILENAME = "lambkin_metadata.yaml"
--- a/src/lambkin/core/ctx/context.py
+++ b/src/lambkin/core/ctx/context.py
@@ -22,6 +22,8 @@ layer and decorators during a run.
 from __future__ import annotations
 
 import datetime
+import logging
+import shutil
 from dataclasses import dataclass
 from pathlib import Path
 from types import SimpleNamespace
@@ -29,7 +31,11 @@ from typing import Any
 
 import yaml
 
-from lambkin.common import defaults
+from lambkin.common import (
+    compute_run_hash,
+    defaults,
+    is_completed,
+)
 from lambkin.core.process.cgroup import (
     find_delegated_cgroup,
     kill_cgroup_tree,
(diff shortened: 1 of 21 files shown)


src/lambkin/core/ctx/context.py
teresa-ortega, line 68: The `shell` attribute is listed in the class docstring without any caveat, but it's only created in `__enter__` on a cache miss, on a cache hit the attribute doesn't exist and accessing it raises `AttributeError`.

test/core/test_benchmark.py
```diff
@@ -398,3 +401,37 @@ def fn(ctx):
 
     fn(args=["--variants", "2"], base_dir=tmp_path)
     assert contexts[0].paths.variant_dir.name == "var_2"
+
+
+def test_benchmark_skips_completed_iterations_on_rerun(variants, tmp_path):
```
teresa-ortega, line 406: `test_benchmark_skips_completed_iterations_on_rerun` verifies that `fn` is never called on the second run, but doesn't assert `ctx.skipped is True` for those iterations. If the skip logic ever moved elsewhere, the test could keep passing for the wrong reasons. Consider collecting the contexts on the second run and asserting all of them have `skipped=True`.

test/core/test_input.py
```diff
@@ -158,11 +158,11 @@ def dataset(ctx):
     def map(ctx):
         return "path/to/map.yaml"
 
-    nominal(base_dir="/tmp")
+    nominal(base_dir=tmp_path)
     assert seen == [("path/to/dataset.mcap", "path/to/map.yaml")]
 
 
-def test_input_hook_called_once_regardless_of_variants_and_iterations():
+def test_input_hook_called_once_regardless_of_variants_and_iterations(tmp_path):
```
teresa-ortega, line 165: No test covers what happens with input hooks when all iterations are cache hits on a second run, inputs are resolved before the loop, so they'd be called regardless of whether any iteration actually runs. Worth adding a test that verifies input hooks are still called (or explicitly documents that they are) even when the entire benchmark is skipped.

src/lambkin/core/ctx/context.py
xaru8145: Done in e8ea48f87af54653b8ec882e8ec2c44404ce9318

test/core/test_input.py
xaru8145: Done in 4858227b79a010dbca660d128d2c614a035b51aa

test/core/test_benchmark.py
xaru8145: fn is never called on skipped iterations so we can't collect skipped contexts from inside fn

xaru8145: Only able to do this 8e1f07bda6d99f847e7a0d52c3f0220833a8d0e9 without complicating too much the current setup

teresa-ortega: You're right , if the iteration is skipped, `fn` is never called so there's no way to collect the context from inside it. Sorry

teresa-ortega (approved):
LGTM

PR #168, opened 2026-06-10
Phase 5: Restore previous terminal state when interactive

xaru8145:
### Proposed changes

The terminal state was not being restored after a `SIGUSR1`-triggered interruption (added in https://github.com/Ekumen-OS/lambkin/pull/162)  killed the benchmark script process mid-run, leaving the terminal in a bad state. This was caused by `start_new_session=True` detaching the script from the terminal's process group — when the script is killed, the terminal's foreground process group is left in an inconsistent state.

The fix saves the terminal attributes before launching the benchmark script and restores them after it exits, in both the normal and `KeyboardInterrupt` paths. Terminal state is restored using `termios.tcgetattr`/`tcsetattr` rather than `stty sane` — this preserves the exact attributes that existed before the benchmark ran, rather than resetting to a generic set of defaults that may differ from the user's configuration.

Additionally, inline comments were expanded to provide better readability and maintainability.

#### How to Test

Add `.values()` to the beluga launch file to make it fail:
```py
    return LaunchDescription(
        [
            config_file_arg,
            map_path_arg,
            laser_model_arg,
            max_particles_arg,
            beluga_node,
            map_server_node,
            lifecycle_manager_node,
        ].values() # add .values() to make it fail
    ) 
```

Run the beluga_benchmark script 

```bash
```

The benchmark should fail with an error. After it exits, verify that the terminal is fully functional — typed characters should appear on screen and commands should behave normally. If the terminal is unresponsive, characters are invisible, or input is not echoed, the terminal state was not restored correctly.

#### Type of change

💥 **Breaking change!** N/A

### Checklist

### Additional comments

Relates to https://github.com/Ekumen-OS/lambkin/pull/162

### DIFF
--- a/src/lambkin/cli.py
+++ b/src/lambkin/cli.py
@@ -29,6 +29,7 @@ import os
 import signal
 import subprocess
 import sys
+import termios
 import uuid
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


teresa-ortega (approved):
LGTM

PR #169, opened 2026-06-11
Phase 5: Introduce scoped context hierarchy (BenchmarkContext, VariantContext, IterationContext)

teresa-ortega:
### Proposed changes

The previous `Context` class handled all three levels of the benchmark loop in a single monolithic object. This made lifecycle hard to reason about, required a hack (`base_ctx`) to resolve inputs before the loop, and blocked future work like scoped inputs or parallel runs.

This PR introduces three composable context classes that map directly to the three `with` levels in the benchmark loop:

* `BenchmarkContext` — owns `source`, `options`, `base_dir`, `variants`, and `variants.yaml` writing. Entered once for the full run.
* `VariantContext` — owns `variant`, `variant_index`, and `variant_dir` creation. Entered once per variant.
* `IterationContext` — replaces `Context`. Owns the cache check, `cgroup` setup, `shell` proxy, output paths, and metadata writes. Delegates `source`, `options`, and `variant` upward via explicit properties.

Also adds `IterationContext.from_params()` for standalone construction without the parent chain.

Input resolution (`inputs.resolve(bctx)`) is called directly in `benchmark.py` — `InputRegistry` belongs to the decorator closure, not to the context.

### How to test
Run beluga benchmark script
```bash
uv run lambkin  examples/beluga/beluga_benchmark.py 
```

#### Type of change

### Checklist

### Additional comments

The three-level hierarchy is the foundation for future scoped inputs/outputs  (scope="variant"), parallel runs, and resource monitoring, all of which require distinct lifecycle boundaries between benchmark, variant, and iteration.


src/lambkin/core/ctx/benchmark_context.py
```diff
@@ -0,0 +1,157 @@
...
+        return path
+
+    @property
+    def paths(self) -> RunPaths:
+        """Minimal RunPaths so input hooks can access ``ctx.paths.base_dir``.
+
+        The variant/iteration indices are fixed at 0 and the derived
+        subdirectories are never created.
+        """
+        return RunPaths.from_indices(self._base_dir, 0, 0)
```
xaru8145: The paths property returns a RunPaths with dummy indices (0, 0) purely so input hooks can access ctx.paths.base_dir. This is a leaky workaround — input hooks shouldn't need ctx.paths.iteration_dir (which doesn't exist at benchmark scope) and the dummy RunPaths makes it look like they could. A dedicated base_dir property is already there, so ctx.paths at benchmark scope is unnecessary and potentially confusing.

src/lambkin/core/ctx/benchmark_context.py
```diff
@@ -0,0 +1,157 @@
...
+            inputs: The ``InputRegistry`` holding hooks registered via
+                ``@benchmark.input``.
+
+        Returns:
+            A ``SimpleNamespace`` with one attribute per registered hook,
+            or ``None`` if no hooks are registered.
+        """
+        return inputs.resolve(self)
+
+    def write_variants_yaml(self, variants: list[dict]) -> Path:
```
xaru8145: Naming suggestion to match `write_metadata()` style, which also writes a yaml. Also, better be private and call within __enter__ rather than have the benchmark call this function. Add comment also in benchmark file.

```suggestion
    def _write_variants(self, variants: list[dict]) -> Path:
```

src/lambkin/core/ctx/iteration_context.py
```diff
@@ -0,0 +1,343 @@
...
+        }
+        if completed_at is not None:
+            metadata["completed_at"] = completed_at
+        with open(self.metadata_path, "w") as f:
+            yaml.dump(metadata, f, default_flow_style=False, sort_keys=False)
+
+    @property
+    def metadata_path(self) -> Path:
+        """Path to the metadata YAML file for this iteration."""
+        return self._paths.iteration_dir / self.METADATA_FILENAME
```
xaru8145: Suggestion: move before `__enter___`  so all properties are defined next to each other and read better.

src/lambkin/core/ctx/iteration_context.py
```diff
@@ -0,0 +1,343 @@
...
+        self._paths = RunPaths.from_indices(
+            variant_ctx.base_dir,
+            variant_ctx.variant_index,
+            iteration,
+        )
+        self._run_hash = compute_run_hash(
+            variant_ctx.raw_variant,
+            iteration,
+            variant_ctx.raw_options,
+        )
```
xaru8145, line 111: `raw_options` and `raw_variant` can be removed — `vars(variant_ctx.variant)` and `vars(variant_ctx.options)` produce equivalent dicts for `compute_run_hash`, and since the hash uses `sort_keys=True` the result is order-independent. This removes the dual representation and avoids leaking internal details as public properties.

src/lambkin/core/ctx/benchmark_context.py
```diff
@@ -0,0 +1,157 @@
...
+
+    @property
+    def base_dir(self) -> Path:
+        """Root directory for all benchmark results."""
+        return self._base_dir
+
+    @property
+    def raw_options(self) -> dict[str, Any]:
+        """Raw options dict (used by child contexts and cache logic)."""
+        return self._raw_options
```
xaru8145: Raw options are not needed. See comment in IterationContext.

src/lambkin/core/ctx/variant_context.py
```diff
@@ -0,0 +1,132 @@
...
+
+    @property
+    def variant_dir(self) -> Path:
+        """Output directory for this variant (e.g. ``results/var_1/``)."""
+        return self._variant_dir
+
+    @property
+    def raw_variant(self) -> dict[str, Any]:
+        """Raw variant dict (used by IterationContext and cache logic)."""
+        return self._raw_variant
```
xaru8145: Raw variant is not needed. See comment in IterationContext.

src/lambkin/core/ctx/variant_context.py
```diff
@@ -0,0 +1,132 @@
...
+
+    @property
+    def base_dir(self) -> Path:
+        """Root benchmark output directory — delegated to BenchmarkContext."""
+        return self._benchmark_ctx.base_dir
+
+    @property
+    def raw_options(self) -> dict:
+        """Raw options dict — delegated to BenchmarkContext."""
+        return self._benchmark_ctx.raw_options
```
xaru8145: Raw options are not needed. See comment in IterationContext.

src/lambkin/core/ctx/variant_context.py
```diff
@@ -0,0 +1,132 @@
...
+from __future__ import annotations
+
+from pathlib import Path
+from types import SimpleNamespace
+from typing import Any
+
+from .benchmark_context import BenchmarkContext
+
+
+class VariantContext:
```
xaru8145, line 35: Missing the `@classmethod def from_params` like done for IterationContext.

src/lambkin/core/ctx/iteration_context.py
```diff
@@ -0,0 +1,343 @@
...
+        if not no_cache and is_completed(self.metadata_path, self._run_hash):
+            self._skipped = True
+            logger.info(
+                "var_%d/iter_%d cache hit, skipping.",
+                self._variant_ctx.variant_index + 1,
+                self._iteration + 1,
+            )
+            return self
+
+        dry_run = getattr(self.options, "dry_run", defaults.DRY_RUN)
```
xaru8145: self.options.dry_run is sufficient — dry_run, no_cache, and log_output are SDK options so they're always present in the parsed options dict. The getattr(..., default) fallback is dead code and can be removed.

src/lambkin/core/ctx/iteration_context.py
```diff
@@ -0,0 +1,343 @@
...
+
+        Args:
+            exc_type: Exception type if propagating, else ``None``.
+            exc_val: Exception value if propagating, else ``None``.
+            exc_tb: Exception traceback if propagating, else ``None``.
+        """
+        if self._skipped:
+            return
+
+        dry_run = getattr(self.options, "dry_run", defaults.DRY_RUN)
```
xaru8145: self.options.dry_run is sufficient — dry_run, no_cache, and log_output are SDK options so they're always present in the parsed options dict. The getattr(..., default) fallback is dead code and can be removed.

src/lambkin/core/ctx/iteration_context.py
```diff
@@ -0,0 +1,343 @@
...
+
+    @property
+    def variant_index(self) -> int:
+        """Zero-based variant position — delegated to VariantContext."""
+        return self._variant_ctx.variant_index
+
+    @property
+    def options(self) -> SimpleNamespace:
+        """Namespaced runtime options — delegated to BenchmarkContext."""
+        return self._variant_ctx.options
```
xaru8145, line 207: The docstring says "delegated to BenchmarkContext" but the property accesses `self._variant_ctx.options`, which itself delegates to `BenchmarkContext`. Update to "delegated to BenchmarkContext via VariantContext" to accurately reflect the chain.

src/lambkin/core/decorators/benchmark.py
xaru8145, line 1: You have removed many code comments that I think can be useful. I agree some of them can be improved or are no longer valid after the change, but modify them rather than eliminating. I think code comments are very valuable for everyone, specially to the people who will be reading and maintaining the code in the future!

src/lambkin/core/decorators/benchmark.py
```diff
@@ -34,12 +34,13 @@
...
 import click
-import yaml
 from click.formatting import HelpFormatter
 
 from lambkin.common import defaults, signals
-from lambkin.core.ctx.context import Context
+from lambkin.core.ctx.benchmark_context import BenchmarkContext
+from lambkin.core.ctx.iteration_context import IterationContext
 from lambkin.core.ctx.source import Source
+from lambkin.core.ctx.variant_context import VariantContext
```
xaru8145: Why not?
```suggestion
from lambkin.core.ctx import (
    BenchmarkContext,
    IterationContext,
    VariantContext,
)
from lambkin.core.ctx.source import Source
```

src/lambkin/core/decorators/benchmark.py
```diff
@@ -218,8 +219,13 @@ def decorator(fn):
...
         def wrapper(args=None, base_dir=None):
-            # Parse CLI arguments, falling back to sys.argv if no args are provided.
-            cli_args = sys.argv[1:] if args is None else args
+            # Parse CLI arguments, falling back to sys.argv if no args provided.
+            if args is not None:
+                cli_args = args
+            elif base_dir is not None:
+                cli_args = []
+            else:
+                cli_args = sys.argv[1:]
```
xaru8145: I don't understand this change as I do not see how base_dir affects cli_args resolution. Is this intentional or could it be a bug?

src/lambkin/core/decorators/benchmark.py
```diff
@@ -262,75 +264,41 @@ def wrapper(args=None, base_dir=None):
...
-                        variant_index=variant_index,
-                    ) as ctx:
-                        if ctx.skipped:
-                            continue
-                        fn(ctx)
-
-            # Calculate and print total elapsed time, useful for user introspection.
+            with BenchmarkContext(source, options, base_dir) as bctx:
+                resolved_inputs = bctx.resolve_inputs(inputs)
+                bctx.write_variants_yaml(variants)
```
xaru8145: What about doing this like in metadata for IterationContext and have `__enter__` call that function and write variants? More self-contained. In this case we can promote both write functions to private, which means add the underscore to the front.

src/lambkin/common/signals.py
xaru8145, line 1: Why this change? It has nothing to do with scoped contexts and I would prefer to review/discuss separately.

xaru8145 (changes requested):
First pass

src/lambkin/core/ctx/iteration_context.py
xaru8145: And make it private (`_metadata_path`) as this is for internal use only.

src/lambkin/core/ctx/benchmark_context.py
teresa-ortega: Solved in 4c4ed47

src/lambkin/core/ctx/benchmark_context.py
teresa-ortega: Solved in e463747

src/lambkin/core/ctx/iteration_context.py
teresa-ortega: Solved in 0b3402f

src/lambkin/core/ctx/iteration_context.py
teresa-ortega: Solved in 556c27f

src/lambkin/core/ctx/benchmark_context.py
teresa-ortega: Solved in 556c27f

src/lambkin/core/ctx/variant_context.py
teresa-ortega: Solved in 556c27f

src/lambkin/core/ctx/variant_context.py
teresa-ortega: Solved in 556c27f

src/lambkin/core/ctx/variant_context.py
teresa-ortega: Solved in 9ed94ea

src/lambkin/core/ctx/iteration_context.py
teresa-ortega: Solved in ba0d636

src/lambkin/core/ctx/iteration_context.py
teresa-ortega: Solved in ba0d636

src/lambkin/core/ctx/iteration_context.py
teresa-ortega: Solved in 877988f

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in ab90310

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in e5afe8c

src/lambkin/core/ctx/benchmark_context.py
```diff
@@ -0,0 +1,122 @@
...
+
+        Args:
+            inputs: The ``InputRegistry`` holding hooks registered via
+                ``@benchmark.input``.
+
+        Returns:
+            A ``SimpleNamespace`` with one attribute per registered hook,
+            or ``None`` if no hooks are registered.
+        """
+        return inputs.resolve(self)
```
xaru8145: `resolve_inputs` doesn't need to be a method on `BenchmarkContext` — it just calls `inputs.resolve(self)`, which `benchmark.py` can do directly: `resolved_inputs = inputs.resolve(bctx)`. `InputRegistry` belongs to the decorator closure, not to the context.

We will see if we need to do something like this for Scoped inputs, but that will come later.

src/lambkin/core/decorators/benchmark.py
teresa-ortega: You're right, this is a bug. base_dir has no relation to CLI argument resolution. This was introduced to avoid pytest injecting its own flags into the Click parser during tests, but the fix was in the wrong place. Reverting to cli_args = sys.argv[1:] if args is None else args and updating the tests to pass args=[] explicitly when no CLI args are needed. Solved in b092bf5

src/lambkin/core/decorators/benchmark.py
teresa-ortega: Solved in 6fa816d

src/lambkin/common/signals.py
teresa-ortega: Solved in 4c4ed47

src/lambkin/core/ctx/benchmark_context.py
teresa-ortega: Solved in 1ed4ad5

src/lambkin/common/signals.py
```diff
@@ -61,14 +61,8 @@ def _handle_sigusr1(signum, frame):
 def setup():
     """Register the SIGUSR1 handler for the benchmark script process.
 
-    Must be called once before any BackgroundProcess is started. Registered
-    in the benchmark script process rather than the CLI because the CLI
```
xaru8145, line 65: You also need to revert the function documentation to what it was

src/lambkin/core/ctx/benchmark_context.py
```diff
@@ -0,0 +1,105 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Benchmark-scoped context.
+
+Owns the outermost lifecycle of a benchmark run: source identity, parsed
+options, base output directory, input resolution, and variants.yaml writing.
```
xaru8145: Variants writting and input resolution are no longer owned by BenchmarkContext.
```suggestion
options and base output directory.
```

src/lambkin/core/ctx/benchmark_context.py
```diff
@@ -0,0 +1,105 @@
...
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+from types import SimpleNamespace
+from typing import Any
+
+from .source import Source
+
+logger = logging.getLogger(__name__)
```
xaru8145: Remove `logger` object and `logging` import as it is no longer used.

src/lambkin/core/ctx/variant_context.py
```diff
@@ -0,0 +1,151 @@
...
+        """Zero-based index of this variant in the full benchmark sweep."""
+        return self._variant_index
+
+    @property
+    def variant_dir(self) -> Path:
+        """Output directory for this variant (e.g. ``results/var_1/``)."""
+        return self._variant_dir
+
+    @property
+    def source(self):
```
xaru8145: Missing return type annotation 
```suggestion
    def source(self) -> Source:
```

src/lambkin/core/ctx/iteration_context.py
```diff
@@ -0,0 +1,337 @@
...
+            variant_index: Zero-based index of this variant.
+            iteration: Zero-based repetition index within this variant.
+            options: Parsed options dict.
+            source: Source object describing the benchmark script.
+            base_dir: Root directory for all benchmark results.
+            inputs: Optional resolved input namespace.
+
+        Returns:
+            A fully initialized ``IterationContext`` ready to be used as a
+            context manager.
```
xaru8145: "A fully initialized" could be confusing as one could think that means __enter__ is executed.
```suggestion
            An ``IterationContext`` ready to be used as a
            context manager.
```

src/lambkin/core/ctx/variant_context.py
```diff
@@ -0,0 +1,151 @@
...
+        Args:
+            variant: Algorithm parameters for this variant.
+            variant_index: Zero-based index of this variant.
+            options: Parsed options dict.
+            source: Source object describing the benchmark script.
+            base_dir: Root directory for all benchmark results.
+
+        Returns:
+            A fully initialized ``VariantContext`` ready to be used as a
+            context manager.
```
xaru8145: "A fully initialized" could be confusing as one could think that means __enter__ is executed.

```suggestion
            A ``VariantContext`` ready to be used as a
            context manager.
```

src/lambkin/core/ctx/variant_context.py
xaru8145: This made me realize that we have mypy on pyproject toml but we never run it, neither via pre-commit nor on a workflow. You can check the mypy issues running `uv run mypy src/`. We can amend later.

test/core/test_benchmark.py
```diff
@@ -122,7 +122,7 @@ def test_benchmark_loops_over_variants_and_iterations(variants, tmp_path):
     def fn(ctx):
         calls.append(ctx)
 
-    fn(base_dir=tmp_path)
+    fn(args=[], base_dir=tmp_path)
```
xaru8145, line 125: Why is `args=[]` needed here as of this PR? And in all other test cases in this file.

test/core/test_context.py
```diff
@@ -12,370 +12,358 @@
...
-    tmp_path, base_variant, base_options, base_source
-):
-    """ctx.skipped is True when the iteration was already completed."""
-    with Context(
-        variant=base_variant,
-        iteration=0,
-        options=base_options,
-        source=base_source,
-        base_dir=tmp_path,
+    def test_skipped_is_true_on_cache_hit(self, tmp_path, source):
```
xaru8145, line 255: Missing test: shell raises AttributeError on cache hit — test_shell_raises_before_enter covers pre-enter access but there's no test for accessing shell when ctx.skipped is True.

test/integration/test_cgroups_cleanup.py
```diff
@@ -37,13 +37,13 @@ def test_cgroups_cleanup(tmp_path):
     child_cgroup = None
     iteration_cgroup = None
 
-    with Context(
+    with IterationContext.from_params(
         variant={},
+        variant_index=0,
         iteration=0,
-        options={},
+        options={"dry_run": False, "no_cache": False, "log_output": "file"},
```
xaru8145: Aren't we missing log_level?

xaru8145 (changes requested):
Minor things, almost there

src/lambkin/common/signals.py
teresa-ortega: Solved in 900f371

src/lambkin/core/ctx/benchmark_context.py
teresa-ortega: Solved in 19a7986

src/lambkin/core/ctx/benchmark_context.py
teresa-ortega: Solved in 19a7986

src/lambkin/core/ctx/variant_context.py
teresa-ortega: Solved in 991cdb2.

src/lambkin/core/ctx/variant_context.py
teresa-ortega: Solved in 991cdb2

src/lambkin/core/ctx/iteration_context.py
teresa-ortega: Solved in 991cdb2

test/core/test_benchmark.py
teresa-ortega: Solved in 6db0581

test/core/test_context.py
teresa-ortega: Solved in 4c434ef

test/integration/test_cgroups_cleanup.py
teresa-ortega: Solved in a9bcc23

xaru8145 (approved):
LGTM! Solid work

PR #170, opened 2026-06-12
Phase 5 : Add lambkin.data package with access.iterations() and evo.series()/stats()

teresa-ortega:
### Proposed changes

Adds the `evo_ape bag2` call to the `nominal` benchmark function, generating `output.ape.zip` so the `plots` and `stats` output hooks have data to consume. Also wires up the `lambkin.data.evo` package end-to-end against a real Beluga AMCL run.

* Calls `ctx.shell.evo_ape.bag2("output", "/ground_truth", "/pose", "--t_max_diff", "0.5", "--save_results", "output.ape.zip")` after the bag playback completes.
* `--t_max_diff 0.5` is required because `evo_ape`'s default pose-matching threshold (`0.01s`) is too tight for this setup: the `/pose` topic published by Beluga has slight timing jitter relative to `/ground_truth`, and after ~50s the accumulated drift between the two clocks exceeds the default threshold, causing evo to discard most pose pairs (only 42 samples were retained). Raising the threshold to `0.5s` (500ms) is still accurate enough for our purposes and recovers the full run — 316 samples covering the complete 132s benchmark.
* `--save_results`  passed as positional arguments rather than kwargs, since `evo` expects underscored flags (`--save_results`) while `ShellProxy` converts kwarg underscores to dashes (`--save-results`).
* Fixed the Beluga sensor topic from `/scan` to `/scan_front` to match the dataset.

### How to test

```bash
uv run lambkin examples/beluga/beluga_benchmark.py --no-cache
```

Expected: each iteration produces an `output.ape.zip`, and the `plots` and `stats` output hooks run once after the loop, saving `examples/beluga/results/plots.png` and logging RMSE/mean/max APE per iteration.
plots

#### Type of change

💥 No breaking changes

### Checklist

### Additional comments

N/A


src/lambkin/data/evo.py
```diff
@@ -0,0 +1,135 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Data access module for evo_ape results.
+
+Provides utilities to collect and aggregate evo_ape result files
+produced by benchmark runs.
+"""
+
+import warnings
```
xaru8145: Let's use `logging` library.

src/lambkin/data/evo.py
```diff
@@ -0,0 +1,135 @@
...
+
+        - ``iter_dir``: :class:`~pathlib.Path` to the iteration directory.
+        - ``variant``: variant directory name.
+        - ``iteration``: iteration index.
+        - ``params``: variant parameters as a :class:`~types.SimpleNamespace`.
+        - ``time``: array of timestamps from start in seconds.
+        - ``ape``: array of APE error values in meters.
+        - ``distance``: array of distances from start in meters (may be None).
+    """
+    file_interface = import_module("evo.tools.file_interface")
```
xaru8145: Explicitly import it on the top. We already depend on evo anyway.

src/lambkin/data/access.py
```diff
@@ -0,0 +1,72 @@
...
+        each with the following attributes:
+
+        - ``iter_dir``: :class:`~pathlib.Path` to the iteration directory.
+        - ``variant``: variant directory name (e.g. ``"var_1"``).
+        - ``iteration``: iteration index (int).
+        - ``params``: :class:`~types.SimpleNamespace` of variant parameters.
+    """
+    root = ctx_or_path.base_dir if not isinstance(ctx_or_path, Path) else ctx_or_path
+    results = []
+    for meta_path in sorted(root.glob("var_*/iter_*/lambkin_metadata.yaml")):
```
xaru8145: Let's use this constant for maintainability:

https://github.com/Ekumen-OS/lambkin/blob/82745300cf49497fb0ef1d18067de00b306f0009/src/lambkin/core/ctx/iteration_context.py#L85

To avoid circular dependencies, you can move it to defaults.

src/lambkin/data/access.py
```diff
@@ -0,0 +1,72 @@
...
+
+@overload
+def iterations(ctx_or_path: Path) -> list: ...
+
+
+@overload
+def iterations(ctx_or_path: object) -> list: ...
+
+
+def iterations(ctx_or_path):
```
xaru8145: I know I suggested this, but let's simplify in favor of Just using a Union type hint and drop the overloads entirely.

xaru8145 (changes requested):
Quick pass

examples/beluga/beluga_benchmark.py
```diff
@@ -77,7 +79,18 @@ def groundtruth(ctx):
...
 def plots(ctx):
     """Generate and save benchmark plots using aggregated evo_ape results."""
-    lambkin.logger.info(f"output called with ctx: {ctx.paths.base_dir}")
+    for entry in lambkin.data.evo.series(ctx, "output.ape.zip"):
+        plt.plot(
+            entry.time, entry.ape, label=f"{entry.variant} / iter {entry.iteration}"
+        )
+
+    for entry in lambkin.data.evo.stats(ctx, "output.ape.zip"):
+        print(f"{entry.variant} iter {entry.iteration}: rmse={entry.rmse:.4f}")
```
xaru8145: What about moving this into a dedicated function?

```py
@nominal.output
def stats(ctx):
    for entry in lambkin.data.evo.stats(ctx, "output.ape.zip"):
        lambkin.logger.info(f"{entry.variant} iter {entry.iteration}: rmse={entry.rmse:.4f}")
```

src/lambkin/data/evo.py
teresa-ortega: Solved in abf5e57

src/lambkin/data/evo.py
```diff
@@ -0,0 +1,137 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Data access module for evo_ape results.
+
+Provides utilities to collect and aggregate evo_ape result files
```
xaru8145: What about other results provided by evo_rpe?

src/lambkin/data/evo.py
teresa-ortega: Solved in 9c54cad

src/lambkin/data/access.py
teresa-ortega: Solved in 91ac8e4

src/lambkin/data/access.py
teresa-ortega: Solved in 03cf06e

examples/beluga/beluga_benchmark.py
teresa-ortega: Solved in 4560ce9

src/lambkin/data/evo.py
teresa-ortega: Solved in c0e2182

src/lambkin/data/evo.py
```diff
@@ -0,0 +1,131 @@
...
+    Returns:
+        A list of :class:`~types.SimpleNamespace` objects, one per iteration,
+        each with the following attributes:
+
+        - ``iter_dir``: :class:`~pathlib.Path` to the iteration directory.
+        - ``variant``: variant directory name.
+        - ``iteration``: iteration index.
+        - ``params``: variant parameters as a :class:`~types.SimpleNamespace`.
+        - ``time``: array of timestamps from start in seconds.
+        - ``ape``: array of APE error values in meters.
```
xaru8145: The field name `ape` is hardcoded in both the docstring and the implementation — ape=result.np_arrays.get("error_array"). Since `series()` is meant to work for any evo metric, not just APE, how would you rename that field so it's generic enough to cover APE, RPE, and any future metric

src/lambkin/data/evo.py
```diff
@@ -0,0 +1,131 @@
...
+            continue
+        result = file_interface.load_res_file(result_path)
+        results.append(
+            SimpleNamespace(
+                iter_dir=entry.iter_dir,
+                variant=entry.variant,
+                iteration=entry.iteration,
+                params=entry.params,
+                time=result.np_arrays.get("seconds_from_start"),
+                ape=result.np_arrays.get("error_array"),
```
xaru8145: ape is hardcoded here

src/lambkin/data/evo.py
```diff
@@ -0,0 +1,131 @@
...
+def log_stats(ctx_or_path: Path | object, filename: str) -> None:
+    """Log evo_ape statistics for all iterations at INFO level.
+
+    Args:
+        ctx_or_path: a benchmark context or a :class:`~pathlib.Path` to the
+            benchmark base directory.
+        filename: name of the evo_ape result zip file (e.g. ``"output.ape.zip"``).
+    """
+    for entry in stats(ctx_or_path, filename):
+        logger.info("%s iter %d: rmse=%.4f", entry.variant, entry.iteration, entry.rmse)
```
xaru8145: Is this the right place for it? Logging is a concern of the benchmark runner, not the data access layer. The user can trivially write this loop themselves, and putting it in `lambkin.data.evo` conflates data access with reporting. It feels like cookbook material rather than SDK.

src/lambkin/data/access.py
```diff
@@ -0,0 +1,65 @@
...
+    for meta_path in sorted(root.glob(f"var_*/iter_*/{defaults.METADATA_FILENAME}")):
+        iter_dir = meta_path.parent
+        variant_name = iter_dir.parent.name
+        with open(meta_path) as f:
+            meta = yaml.safe_load(f)
+        results.append(
+            SimpleNamespace(
+                iter_dir=iter_dir,
+                variant=variant_name,
+                iteration=meta.get("iteration", 0),
```
xaru8145: defaulting to 0 silently hides a missing or malformed metadata file. If iteration is absent it means something went wrong during the run, and you want to know about it, not silently get 0. Should be `meta["iteration"]` to fail loudly.

src/lambkin/data/access.py
```diff
@@ -0,0 +1,65 @@
...
+                params=SimpleNamespace(**meta.get("variant", {})),
```
xaru8145: Same comment here as the iteration.

src/lambkin/data/access.py
```diff
@@ -0,0 +1,65 @@
...
+            )
+        )
```
xaru8145, line 74: Also worth checking `completed_at` here — if the metadata has no completed_at it means the iteration didn't finish successfully, and `iterations()` should warn or skip it rather than silently including it:

src/lambkin/data/access.py
```diff
@@ -0,0 +1,69 @@
...
+        - ``iteration``: iteration index (int).
+        - ``params``: :class:`~types.SimpleNamespace` of variant parameters.
+    """
+    root = (
+        ctx_or_path.base_dir  # type: ignore[attr-defined]
+        if not isinstance(ctx_or_path, Path)
+        else ctx_or_path
+    )
+    results = []
+    for meta_path in sorted(root.glob(f"var_*/iter_*/{defaults.METADATA_FILENAME}")):
```
xaru8145: It works but `sorted()` on paths sorts lexicographically, so `var_10` would come before `var_2`. Is that what we want?

src/lambkin/data/access.py
```diff
@@ -0,0 +1,69 @@
...
+        iter_dir = meta_path.parent
+        variant_name = iter_dir.parent.name
```
xaru8145: Why not get `variant_name` from the metadata?

xaru8145 (changes requested):

src/lambkin/data/access.py
```diff
@@ -0,0 +1,69 @@
...
+
+from pathlib import Path
+from types import SimpleNamespace
+
+import yaml
+
+from lambkin.common import defaults
+
+
+def iterations(ctx_or_path: Path | object) -> list:
```
xaru8145: What about accepting also the path as a string? Same for evo.py

xaru8145: Also, encoding the possible types as the input name seems weird. Would rename to `source` or `target` or other

src/lambkin/data/evo.py
teresa-ortega: Solved in 5306cae

src/lambkin/data/evo.py
teresa-ortega: Solved in 5306cae

src/lambkin/data/evo.py
teresa-ortega: Solved in 74fada3

src/lambkin/data/access.py
teresa-ortega: Solved in c6fb751

src/lambkin/data/access.py
teresa-ortega: Solved in c6fb751

src/lambkin/data/access.py
teresa-ortega: Solved in 8a5fe40

src/lambkin/data/access.py
teresa-ortega: Solved in df04fe2

src/lambkin/data/access.py
teresa-ortega: Solved in 5e01ae7

src/lambkin/data/access.py
teresa-ortega: Solved in 3f550b6

examples/beluga/beluga_benchmark.py
```diff
@@ -74,10 +76,23 @@ def groundtruth(ctx):
...
 @nominal.output
 def plots(ctx):
     """Generate and save benchmark plots using aggregated evo_ape results."""
-    lambkin.logger.info(f"output called with ctx: {ctx.paths.base_dir}")
+    _log_stats(ctx, "output.ape.zip")
+
+    plt.xlabel("Time (s)")
+    plt.ylabel("APE (m)")
+    plt.legend()
+    plt.savefig(ctx.base_dir / "plots.png")
```
xaru8145: Review this please, it is not correct. plots will not plot anything. Why have a hidden function for logging? Use @nominal.output to register another output to print the stats. Grountruth nominal input is no longer needed

src/lambkin/data/access.py
xaru8145, line 1: Have you considered returning an iterator for lazy evaluation rather than the entire list? Like old lambkin

src/lambkin/data/access.py
```diff
@@ -0,0 +1,77 @@
...
+from lambkin.common import defaults
+
+logger = logging.getLogger(__name__)
+
+
+def iterations(source: Path | str | object) -> list:
+    """Traverse the benchmark output tree and return all iteration entries.
+
+    Accepts either a benchmark context (exposing ``paths.base_dir``) or a
+    plain :class:`~pathlib.Path` to the benchmark base directory, so that
```
xaru8145: Update docu to say it also accepts a string. Or you can drop this part since it is already covered in Args. But update Args as well with the str option. In all the files not just this one.

test/data/test_access.py
```diff
@@ -0,0 +1,108 @@
...
+def test_iterations_empty_dir_returns_empty_list(tmp_path):
+    """iterations() returns an empty list when no iteration dirs exist."""
+    assert access.iterations(tmp_path) == []
+
+
+def test_iterations_are_sorted(benchmark_dir):
+    """iterations() returns entries in sorted path order."""
+    entries = access.iterations(benchmark_dir)
+    paths = [e.iter_dir for e in entries]
+    assert paths == sorted(paths)
```
xaru8145: Does this check that those are sorted in the way we want?

xaru8145 (changes requested):

examples/beluga/beluga_benchmark.py
teresa-ortega: Solved in b6f81cf

src/lambkin/data/access.py
teresa-ortega: Considered it, but I think a list is safer here. The typical use case consumes all results at once (plotting, logging), and a one-shot iterator would be a silent bug if someone iterates twice. The numpy arrays are already in memory once the zip is opened, so the memory savings are minimal. We can revisit in Phase 6 when we tackle report generation.

src/lambkin/data/access.py
teresa-ortega: Solved in 472bf74

test/data/test_access.py
teresa-ortega: Solved in 6770993

examples/beluga/beluga_benchmark.py
xaru8145, line 1: How does this even work? We are not saving evo ape results in a zip file. Not sure when this was removed... or if it even was there at some point

xaru8145 (changes requested):

examples/beluga/beluga_benchmark.py
teresa-ortega: Solved in 555d647

xaru8145 (approved):
LGTM

PR #173, opened 2026-06-15
Phase 5: Add non-strict mypy check as part of pre-commit and solve current issues in src/

xaru8145:
### Proposed changes

Prior to this change, `mypy` was configured in `pyproject.toml` but never enforced — it was neither wired into the `pre-commit hooks` nor gated in CI. As a result, type errors had accumulated silently in `src/` over time despite the configuration being present. 

This PR adds `mypy` to `.pre-commit-config.yaml`, making it part of the standard pre-commit pipeline which is already run in CI via `action-pre-commit-uv`. 

Scoping decisions:

- Strict mode is intentionally disabled for now — strictness will be increased incrementally in future phases.
- `tests/` is excluded from the mypy scope to avoid noise and keep the focus on source code.
- With these two decisions, all type errors present in `src/` under non-strict mode have been fixed as part of this PR.

Why a `local` hook instead of `mirrors-mypy`?

- `mirrors-mypy` runs in an isolated environment and cannot see the project's `uv`-managed virtualenv.
- Dependencies would need to be duplicated in `additional_dependencies`, creating a maintenance burden.
- A local hook using `uv run mypy` runs inside the project's own virtualenv with all dependencies available.

Why `pass_filenames: false`?

- Pre-commit by default passes only staged files to the hook, causing mypy to check files in isolation.
- This would miss cross-file type errors
- With `pass_filenames: false`, mypy uses its own file discovery via `pyproject.toml` and checks the full `src`/ tree on every run

#### How to Test

Run mypy:

```
uv sync
uv run mypy
```

Ensure all tests pass:

```
uv run pytest
```

#### Type of change

💥 **No Breaking change!**

### Checklist

### Additional comments

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,3 +24,12 @@ repos:
       - id: ruff
         args: [ --fix ]
       - id: ruff-format
+
+  - repo: local
+    hooks:
+      - id: mypy
+        name: mypy
+        entry: uv run mypy
+        language: system
+        types: [python]
+        pass_filenames: false
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,11 +35,13 @@ dev = [
     "ruff>=0.15.0",
     "mypy>=1.19.1",
     "pytest>=8.0",
+    "types-pyyaml>=6.0.12.20260518",
 ]
 
 [tool.mypy]
 strict = true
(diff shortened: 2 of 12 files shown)


teresa-ortega (approved):
LGTM

PR #177, opened 2026-06-17
Phase 5: Update Documentation 

teresa-ortega:
### Proposed changes

Rewrite and restructure the project's three READMEs (root, SDK, and Beluga example) to match the current state of next-gen.

Remove unused files and folders left over from previous iterations.

#### Type of change

💥 No breaking changes

### Checklist

### Additional comments

N/A


README.md
```diff
@@ -11,21 +11,143 @@ Most benchmarking systems are built around a specific algorithm, dataset format,
...
+| **Result access** | Read structured benchmark outputs (`lambkin.data`) from a notebook or a standalone script, with no live benchmark required. |
 
 To understand how LAMBKIN works under the hood, see the [SDK documentation](src/lambkin/README.md).
 
 
+## Prerequisites
+
+- Python 3.10+
+- [`uv`](https://github.com/astral-sh/uv)
+- Linux with cgroups v2 and systemd (required for background process management)
```
xaru8145: I do not think we require systemd (that was before when we were running processes via systemd run ...). Linux with cgroups v2 is sufficient.

README.md
```diff
@@ -11,21 +11,143 @@ Most benchmarking systems are built around a specific algorithm, dataset format,
...
+        "output.ape.zip",
+    )
+
+
+@my_benchmark.input
+def dataset(ctx):
+    return ctx.source.path.parent / "datasets" / "my_dataset.mcap"
+
+
+@my_benchmark.ouput
```
xaru8145:
```suggestion
@my_benchmark.output
```

README.md
```diff
@@ -11,21 +11,143 @@ Most benchmarking systems are built around a specific algorithm, dataset format,
...
+if __name__ == "__main__":
+    my_benchmark()
+```
+
+Run it with the CLI:
+
+```bash
+lambkin my_benchmark.py
+```
+Run `lambkin my_benchmark.py --help` to see all available options.
```
xaru8145: I think is better without using the script here. Just general sdk usage. The help will already redirect towards how to derive the options registered in the script.
```suggestion
Run `lambkin --help` to see all available options.
```

README.md
```diff
@@ -11,21 +11,143 @@ Most benchmarking systems are built around a specific algorithm, dataset format,
...
+## Use Cases
+
+### Benchmarking different datasets
+
+When the input itself depends on the variant — e.g. a different sensor model needs a different recording — scoping the hook to `"variant"` means it's only resolved when the variant changes, not on every iteration. This keeps a multi-dataset sweep just as cheap as a single-dataset one.
+
+```python
+@nominal.input(scope="variant")
+def dataset(ctx):
+    return ctx.source.path.parent / "datasets" / f"{ctx.variant.sensor_model}.mcap"
```
xaru8145:
```suggestion
@lambkin.benchmark(
    variants=lambkin.common.named_product(
        dataset=["warehouse", "office"]
    ),
    num_iterations=2,
)
def nominal(ctx):
    ...

@nominal.input(scope="variant")
def dataset(ctx):
    return ctx.source.path.parent / "datasets" / f"{ctx.variant.dataset}.mcap"
```

README.md
```diff
@@ -11,21 +11,143 @@ Most benchmarking systems are built around a specific algorithm, dataset format,
...
+Run it with the CLI:
+
+```bash
+lambkin my_benchmark.py
+```
+Run `lambkin my_benchmark.py --help` to see all available options.
+
+For a complete, working example using the Beluga algorithm, see [`examples/beluga/beluga_benchmark.py`](examples/beluga/beluga_benchmark.py). For the full CLI reference, see the [SDK documentation](src/lambkin/README.md#cli).
+
+## Use Cases
```
xaru8145, line 122: Missing more use cases like benchmarking different algorithms

README.md
```diff
@@ -11,21 +11,143 @@ Most benchmarking systems are built around a specific algorithm, dataset format,
...
+
+When the input itself depends on the variant — e.g. a different sensor model needs a different recording — scoping the hook to `"variant"` means it's only resolved when the variant changes, not on every iteration. This keeps a multi-dataset sweep just as cheap as a single-dataset one.
+
+```python
+@nominal.input(scope="variant")
+def dataset(ctx):
+    return ctx.source.path.parent / "datasets" / f"{ctx.variant.sensor_model}.mcap"
+```
+
+## Cookbook
```
xaru8145: Move this out of Results — it doesn't belong there. The snippet uses ctx.shell, which only exists inside the benchmark function during execution (output hooks get a BenchmarkContext with no shell, see the existing warning above). So this isn't a "read back finished results" recipe, it's "convert what this iteration's process just produced into a format evo can score" — that's the Execution moment, not Result access.

Given that, I'd make this a standalone top-level Cookbook section near the end of the SDK README (sibling to Results, not nested under it).

Also: frame it as wrapping evo's own multi-format support rather than a lambkin-specific converter. Maybe mention all format conversions supported and link the evo docu: https://github.com/MichaelGrupp/evo/wiki/Formats#saving--exporting-to-other-formats

README.md
```diff
@@ -11,21 +11,143 @@ Most benchmarking systems are built around a specific algorithm, dataset format,
 
 LAMBKIN separates the orchestration machinery from the benchmark definition. The algorithm runs as an external process — LAMBKIN does not need to know what is inside it. Parameter sweeps, process lifecycle, I/O, and metric collection are all handled by the SDK, so your script stays focused on the benchmark logic.
 
+SLAM algorithms are rarely deterministic — particle filters resample randomly, sensor noise varies run to run, and timing jitter between processes can shift outcomes on its own. A single run's accuracy metric is one sample from a noisy distribution, not a reliable estimate of how an algorithm actually performs. LAMBKIN treats `num_iterations` as a first-class part of every benchmark definition, not an afterthought, so a variant is always evaluated as a population of runs you can summarize — with a mean, a standard deviation, an RMSE — instead of a single number taken on faith.
```
xaru8145: I think paragraph "SLAM algorithms are rarely deterministic..." can be summarized and added into the table below as a capability, instead of having as a paragraph here. For example:

**Iterations** | Run each variant across multiple iterations, giving results statistical significance instead of relying on a single noisy sample.

README.md
```diff
@@ -11,21 +11,143 @@ Most benchmarking systems are built around a specific algorithm, dataset format,
...
+
+The same reasoning extends to where results land on disk. A benchmark only stays reproducible and reportable if its output follows a structure you didn't have to invent for that particular run: `results/var_<n>/iter_<n>/`, with metadata recorded alongside whatever artifacts your commands produced. That predictability is what lets `lambkin.data` walk any benchmark's results without being told their shape in advance, and what lets a notebook, a report generator, or a teammate's script reuse the same data months later without reverse-engineering a one-off layout.
```
xaru8145: Trim this paragraph — it's mixing philosophy with implementation detail it doesn't need. The literal `results/var_<n>/iter_<n>/` pattern already has a home with the actual tree diagram in the SDK README's "Expected Output" section, and naming `lambkin.data` here duplicates the Capabilities table's "Result access" row. This section should just make the argument — predictable structure enables later reuse — without committing to the specific shape or API that backs it:

```suggestion
A benchmark's output deserves the same discipline as its execution: a structured, predictable layout, with metadata recorded alongside whatever artifacts your commands produce. That predictability is what lets a notebook, a report generator, or a teammate's script revisit any benchmark's output months later without reverse-engineering its shape.
```

src/lambkin/README.md
```diff
@@ -2,49 +2,91 @@
...
+- [Process Management with cgroups v2](#process-management-with-cgroups-v2)
+- [CLI](#cli)
+- [Partial Restarts](#partial-restarts)
+- [Logging](#logging)
+- [Results](#results)
+- [Expected Output](#expected-output)
+
 ## Architecture
 
-A benchmark is structured around three stages that LAMBKIN sequences and keeps organized:
```
xaru8145, line 7: Revert two word choices: 
- Setup/Execution/Result access → Ingestion/Execution/Egression (matches the design doc, and "ingestion" covers this cleanly — resolving any dependency, whether it's a dataset file or a generated seed, is still pulling a value in from outside the calling code).
-  "three moments" → "three stages" (the very next sentence already says "Each stage is a plain Python function," so it's currently inconsistent with itself).

src/lambkin/README.md
```diff
@@ -2,49 +2,91 @@
...
-**Input**
+### Input
 
-Registers a data resolution hook on a benchmark function. Each hook must be a callable that accepts a single `Context` object as its argument. Hook names must be unique — registering two hooks with the same name raises an error. Hooks are resolved once before the execution loop and their return values injected into the context under the hook's function name. Can be used as a decorator via `@input`.
+Registers a data resolution hook on a benchmark function. Each hook must be a callable that accepts a single `Context` object as its argument and must return a non-empty value. Hook names must be unique across the entire benchmark — registering two hooks with the same name, even under different scopes, raises an error. Can be used as a decorator via `@input`.
+
+Hooks can be scoped to one of three lifecycle levels via `scope=`:
+
+- `"benchmark"` (default) — resolved once before the variant loop. Use for inputs that don't depend on the current variant or iteration, e.g. a shared dataset.
+- `"variant"` — resolved once per variant. Use for inputs that depend on `ctx.variant` but not `ctx.iteration`, e.g. selecting a dataset file by sensor model.
```
xaru8145: Selecting a dataset per sensor model does not seem a very clear example. Rephrase to selecting a dataset registered as a variant

src/lambkin/README.md
```diff
@@ -2,49 +2,91 @@
...
 
-**Input**
+### Input
 
-Registers a data resolution hook on a benchmark function. Each hook must be a callable that accepts a single `Context` object as its argument. Hook names must be unique — registering two hooks with the same name raises an error. Hooks are resolved once before the execution loop and their return values injected into the context under the hook's function name. Can be used as a decorator via `@input`.
+Registers a data resolution hook on a benchmark function. Each hook must be a callable that accepts a single `Context` object as its argument and must return a non-empty value. Hook names must be unique across the entire benchmark — registering two hooks with the same name, even under different scopes, raises an error. Can be used as a decorator via `@input`.
+
+Hooks can be scoped to one of three lifecycle levels via `scope=`:
+
+- `"benchmark"` (default) — resolved once before the variant loop. Use for inputs that don't depend on the current variant or iteration, e.g. a shared dataset.
```
xaru8145: Nit: better say a single dataset for the entire benchmark. shared does not read that natural to me.

src/lambkin/README.md
```diff
@@ -92,26 +134,66 @@ LAMBKIN is a benchmarking SDK for robotics applications. It runs your
...
+
+`run_hash` is derived from the variant parameters, the iteration index, and the script's custom options. SDK-level flags (`--log-level`, `--dry-run`, and similarly inconsequential ones) are excluded, since they don't affect benchmark outputs. Changing any variant parameter or custom option invalidates only the affected iterations' hashes — and therefore only reruns those — leaving everything else alone.
+
+Pass `--no-cache` to bypass this check entirely and force a full rerun of every iteration, regardless of prior completion.
+
+> [!NOTE]
+> Iterations run with `--dry-run` still write the initial `lambkin_metadata.yaml`, but never get `completed_at` — dry-run iterations don't do real work, so they're never considered complete. Re-running a dry-run script will show no cache hits every time, and will keep overwriting the same incomplete metadata file — that's expected, not a bug.
```
xaru8145: Better simplify
```suggestion
> [!NOTE]
> Dry runs write the initial metadata but never get `completed_at`, so they always show as cache misses — that's expected, not a bug.
```

src/lambkin/README.md
```diff
@@ -92,26 +134,66 @@ LAMBKIN is a benchmarking SDK for robotics applications. It runs your
...
+# List every variant with its var_N number, without running anything.
+lambkin my_benchmark.py --show-variants
+
+# Run only variants 1, 3, 4 and 5.
+lambkin my_benchmark.py --variants 1,3:5
+```
+
+Variant numbers always refer to the full, original sweep — they don't shift when you select a subset, so the output folder for a given configuration (`var_N`) stays stable across runs.
+
+## Partial Restarts
```
xaru8145, line 212: I think this section is a bit too descriptive and we should aim at simplifying and go straight to the point.

Suggestion (pass no cache part is fine):

LAMBKIN supports partial restarts by writing a metadata file to each iteration's own output folder. It records whether the iteration completed, along with a hash encoding the variant's parameters, the iteration index, and any custom options you've declared — SDK flags like `--log-level` are excluded, since they don't affect what the benchmark actually produces.

> [!WARNING]
> Currently the hash doesn't account for changes to the benchmark function's own source code, or to the contents of a file an input hook resolves (e.g. swapping in a different dataset without changing any variant or option). Edit either of those and the cache will still report a hit.

On the next run, before touching anything, LAMBKIN checks each iteration's metadata against the current hash: a match with a recorded completion means it's safe to skip; a mismatch — because a parameter or option changed — or no completion at all means it reruns. That's what lets an interrupted hundred-run sweep resume exactly where it left off, and lets changing one parameter invalidate only the iterations it actually affects.

Pass `--no-cache` to bypass this check entirely and force a full rerun of every iteration, regardless of prior completion.

src/lambkin/README.md
```diff
@@ -121,11 +203,31 @@ lambkin my_benchmark.py --log-level debug
 lambkin my_benchmark.py --log-level DEBUG  # equivalent
 ```
 
-The LAMBKIN logger is fully isolated from the root logger — user scripts can configure their own logging without any interference.
+The LAMBKIN logger is fully isolated from the root logger — user scripts can configure their own logging without any interference. `configure_logging` attaches a dedicated `StreamHandler` to the `"lambkin"` logger and sets `propagate = False`, so its messages never reach the root logger or any handler your script may have configured there. Most scripts don't need to touch this.
```
xaru8145, line 239: On SDK Logging:

- The propagate example doesn't work on its own, and that's not obvious from how it's written. Setting `logging.getLogger("lambkin").propagate = True` only does something if there's already a handler configured on the root logger — and LAMBKIN deliberately never configures root itself (confirmed in `configure_logging`, it only ever touches `"lambkin"`). As written, a reader copying just that one line gets a no-op, since propagation just hands the record to an empty root logger with nothing listening. Either show the full pairing (e.g. `logging.basicConfig(...)` alongside the `propagate = True` line) or say explicitly that this only matters if you're already configuring root yourself.

- Missing the simpler option. If someone just wants LAMBKIN's output to also go to a file, they don't need to touch propagation at all — attaching a handler directly to `"lambkin"` (`logging.getLogger("lambkin").addHandler(...)`) works regardless of propagate, since a logger's own handlers always run independent of that flag. Worth presenting that as the first/default option, with the propagate route as the secondary one for folding LAMBKIN's output into an existing root-level setup specifically.

src/lambkin/README.md
```diff
@@ -152,74 +254,80 @@ Precedence (highest to lowest):
...
-```bash
-git clone -b next-gen git@github.com:Ekumen-OS/lambkin.git
-uv sync
-```
+### Output Hooks
 
-## Usage
-
-A minimal example composing the SDK primitives described above into a working benchmark.
+`@my_benchmark.output` hooks run once, after the entire benchmark loop finishes, against the full results tree — the natural place to call `lambkin.data` and turn it into plots or logged statistics:
```
xaru8145: Trim the opening line of the Results > Output Hooks section — it restates "runs once, after the entire benchmark loop completes," which is already covered by the Output entry in Core Concepts

src/lambkin/README.md
```diff
@@ -152,74 +254,80 @@ Precedence (highest to lowest):
 2. **Per-call** — `log_output` keyword at the call site
 3. **ShellProxy default** — `ShellProxy(log_output="file")`
 
-## Requirements
+## Results
```
xaru8145, line 299: Missing a `### Metrics` section within results. Sth like:

LAMBKIN doesn't compute trajectory metrics itself — it invokes evo through ctx.shell, the same way it invokes any other process, and reads back whatever evo writes. The field names exposed below are evo's own, not LAMBKIN's.

Maybe mention the general use cases like APE and RPE and bind the official evo docu.

src/lambkin/README.md
```diff
@@ -152,74 +254,80 @@ Precedence (highest to lowest):
...
+import lambkin
 
-```bash
-lambkin my_benchmark.py --show-options
-```
 
-Validate the benchmark pipeline without executing any process:
+def reprocess():
+    for it in lambkin.data.access.iterations(
+        "/home/xaru/ekumen/lambkin/examples/beluga/results"
```
xaru8145: No need to refer a path from my computer 😆 Use a generic "/path/to/results" please

src/lambkin/README.md
```diff
@@ -152,74 +254,80 @@ Precedence (highest to lowest):
...
 
-```bash
-lambkin my_benchmark.py --dry-run
+
+if __name__ == "__main__":
+    reprocess()
 ```
 
-For a complete, working example using the Beluga algorithm, see [`Beluga Example`](examples/beluga/beluga_benchmark.py).
+This is the same code as in an `@output` hook, just pointed at a path string instead of `ctx`. Useful for generating a new report from an old run, comparing two separate `results/` directories, or trying out a plot before committing it to the benchmark script itself.
```
xaru8145: Also mention the problem with logging here (while reprocessing). We are not tracking call counts (no need to explain this) so if we run a process in file log mode with the same process, the log file will be overwritten. So we live with that or set log to console. Also, cgroups need to be None as only context manages its lifecycle so not sage to assign a crgoup here.

src/lambkin/README.md
```diff
@@ -152,74 +254,80 @@ Precedence (highest to lowest):
...
 
 ## Expected Output
```
xaru8145: Mention that two lambkin invocations against the same `results/` directory aren't guarded against. So either remove or rename the folder, or place the benchmark in a different folder.

It is a design choice to have reproducible access. If the user could set any path in the results folder we would likely loose cache access if the user forgets to set the same output. And also benchmark scripts should be kind of linked to a given static setup.

src/lambkin/README.md
```diff
@@ -2,49 +2,91 @@
...
 
-**Shell**
+### Shell
 
 Abstracts shell command dispatch. Exposes the host environment's executables as Python attributes — accessing `shell.my_tool` returns a callable that runs `my_tool` with the given arguments, letting benchmark scripts invoke external processes without hardcoding paths or constructing subprocess calls manually. Accessible through the context.
 
-**Background Process**
+### Background Process
 
 Runs a process in the background while the benchmark continues executing. Takes a shell command (without calling it) and manages its full lifecycle — start, monitor, and clean up — as a context manager. When the context exits, it terminates the process and all its descendants; any process spawned outside a `background()` block is not covered.
```
xaru8145, line 122: Explain that we interrupt the benchmark when we detect a failed background process. For example:

"If the background process dies unexpectedly before the context exits, any foreground call currently blocked waiting on it is unblocked immediately instead of hanging, and the iteration fails with `LambkinProcessDiedUnexpectedlyError`."

src/lambkin/README.md
xaru8145: Is this really true in a containerized environment? I think it is delegated to the cgroup root not the suer service.

xaru8145, line 157: Because the current section only explains the mechanical cgroup-kill side of Ctrl-C, not the consequence that actually matters to someone writing a benchmark function: their script's process never receives SIGINT at all, since the CLI deliberately launches it in its own session to shield it. That's a real gap — someone could write a try/except KeyboardInterrupt expecting it to run cleanup logic on interrupt, and it would silently never fire, with nothing in the docs to explain why.

I suggest you add sth this to the existing paragraph on process cleanup: " Your benchmark function's process never receives SIGINT directly — only the CLI's own session does."

xaru8145 (changes requested):
First comprehensive pass. Did not review examples README.

README.md
```diff
@@ -11,21 +11,143 @@ Most benchmarking systems are built around a specific algorithm, dataset format,
...
 To understand how LAMBKIN works under the hood, see the [SDK documentation](src/lambkin/README.md).
 
 
+## Prerequisites
+
+- Python 3.10+
+- [`uv`](https://github.com/astral-sh/uv)
+- Linux with cgroups v2 and systemd (required for background process management)
+
+## Installation
```
xaru8145, line 38: Suggestion for the Installation after carefully thinking about this:

---

```bash
git clone -b next-gen git@github.com:Ekumen-OS/lambkin.git
cd lambkin
uv tool install .
```

This installs lambkin on your `PATH` — you can delete the cloned directory afterward, and rerun this command to pick up updates. Use `--editable` instead if you want it to track your local checkout live (then keep the clone in place).

Alternatively, you can install it into a uv-managed virtual environment:

```bash
uv sync
```

Then run benchmarks via `uv run lambkin my_benchmark.py`.

---

xaru8145 (commented):
Some things I noticed while reviewing:
- `bag2tum.py` is still in `utils/`. We should remove it as we rely on evo for conversion utilities.
- `layers/` folder is unusued, let's remove it.

README.md
teresa-ortega: Solved in 0e1efb7

README.md
teresa-ortega: Solved 0e1efb7

README.md
teresa-ortega: Solved in 0e1efb7

README.md
teresa-ortega: Solved in 0e1efb7

README.md
teresa-ortega: Solved in 1a49bb7

README.md
teresa-ortega: Solved in 1a49bb7

README.md
teresa-ortega: Solved in 23cef14

README.md
teresa-ortega: Solved in 24b200e

src/lambkin/README.md
teresa-ortega: Solved in e2d8d1f

src/lambkin/README.md
teresa-ortega: Solved in 4eb3218

src/lambkin/README.md
teresa-ortega: Solved in 4eb3218

src/lambkin/README.md
teresa-ortega: Solved in 8929c18

src/lambkin/README.md
teresa-ortega: Solved in 7c3fda6

src/lambkin/README.md
teresa-ortega: Solved in 1b29b12

src/lambkin/README.md
teresa-ortega: Solved in b53dfa0

src/lambkin/README.md
teresa-ortega: Solved in 6600e97

src/lambkin/README.md
teresa-ortega: Solved

src/lambkin/README.md
teresa-ortega: Solved in aeeb06d

src/lambkin/README.md
teresa-ortega: Solved in 7903729

src/lambkin/README.md
teresa-ortega: Solved in 0451503

src/lambkin/README.md
teresa-ortega: Solved in 6109484

teresa-ortega: Solved in 3160a9b

README.md
teresa-ortega: Solved in a74d6d1

README.md
```diff
@@ -11,21 +11,160 @@ Most benchmarking systems are built around a specific algorithm, dataset format,
...
+
+
+@my_benchmark.input
+def dataset(ctx):
+    return ctx.source.path.parent / "datasets" / "my_dataset.mcap"
+
+
+@my_benchmark.output
+def plot(ctx):
+    data = lambkin.data.evo.series(ctx)
```
xaru8145: Here it says plot but we do not plot. Might we worth to add the plot command just to have a more representative example.

examples/beluga/README.md
```diff
@@ -129,14 +136,38 @@ podman-compose --profile production run --podman-run-args="--systemd=always" --r
...
 
 > [!WARNING]
-> Both runtimes require elevated privileges to support background process management. Docker runs with --privileged, granting the container broad access to host devices and kernel interfaces. Podman uses --systemd=always, which allows the container to interact with the host's cgroup v2 hierarchy.
+> Both runtimes require elevated privileges to support background process management. Docker runs with `--privileged`, granting the container broad access to host devices and kernel interfaces. Podman uses `--systemd=always`, which allows the container to interact with the host's cgroup v2 hierarchy.
 
-## Usage
+### Run on the Host
 
-Inside the Docker container/enviroment, run the following command:
+If you'd rather run the example directly on the host — e.g. you already have ROS 2 Jazzy and `evo` installed locally — you can extract the bundled reference dataset from the dataset image without starting it:
```
xaru8145: you do not need evo locally. That is included in the dependencies lambkin installs in the virtual environment.

src/lambkin/README.md
```diff
@@ -65,9 +135,9 @@ app.slice/                               ← user's systemd app slice
         └── my_recorder-e3f4a5b6/        ← ros2 bag record process
 ```
 
-When running on the host, a user systemd app slice (app.slice) is always available. In containerized environments no app slice may exist — in that case, LAMBKIN falls back to the nearest delegated cgroup it can find. For example, under Podman with `--systemd=always`:
+When running on the host, a user systemd app slice (app.slice) is always available. In containerized environments no app slice may exist — in that case, LAMBKIN falls back to the nearest delegated cgroup it can find. For example, under Podman with s`--systemd=always`:
```
xaru8145:
```suggestion
When running on the host, a user systemd app slice (app.slice) is always available. In containerized environments no app slice may exist — in that case, LAMBKIN falls back to the nearest delegated cgroup it can find. For example, under Podman with `--systemd=always`:
```

src/lambkin/README.md
xaru8145: I think this has been tackled in the wrong section @teresa-ortega. Seems like the preview can be a bit confusing. This was a comment for the expected output section!

README.md
```diff
@@ -11,21 +11,159 @@ Most benchmarking systems are built around a specific algorithm, dataset format,
...
+
+@nominal.input(scope="variant")
+def dataset(ctx):
+    return ctx.source.path.parent / "datasets" / f"{ctx.variant.dataset}.mcap"
+```
+
+### Benchmarking different algorithms
+
+`named_product` doesn't care whether a parameter is a tuning knob or a completely different code path. Adding an `algorithm` dimension to the sweep is enough to compare two localizers under identical conditions — same dataset, same number of iterations, same evaluation pipeline — with no extra orchestration:
+```python
```
xaru8145, line 159: Spacing (new line).

```suggestion
`named_product` doesn't care whether a parameter is a tuning knob or a completely different code path. Adding an `algorithm` dimension to the sweep is enough to compare two localizers under identical conditions — same dataset, same number of iterations, same evaluation pipeline — with no extra orchestration:

```python
```

README.md
xaru8145: Missing the closing line "Then run benchmarks via `uv run lambkin my_benchmark.py`." I suggested to add. It is relevant, so please do not forget to add it @teresa-ortega

README.md
```diff
@@ -11,21 +11,159 @@ Most benchmarking systems are built around a specific algorithm, dataset format,
...
+@lambkin.benchmark(
+    variants=lambkin.common.named_product(
+        algorithm=["beluga", "nav2_amcl"],
+        sensor_model=["likelihood", "beam"],
+        num_particles=[1, 10, 100, 1000, 10000],
+    ),
+    num_iterations=30,
+)
+@lambkin.option("--clock-rate", default=100.0)
+@lambkin.option("--clock", default=True)
```
xaru8145: I do not think we need to register options here. This is just a simple code snippet to showcase how to benchmark differents algortihms. Options usage is well documented.

README.md
```diff
@@ -11,21 +11,159 @@ Most benchmarking systems are built around a specific algorithm, dataset format,
...
+| **Result access** | Read structured benchmark outputs (`lambkin.data`) from a notebook or a standalone script, with no live benchmark required. |
 
 To understand how LAMBKIN works under the hood, see the [SDK documentation](src/lambkin/README.md).
 
 
+## Prerequisites
+
+- Python 3.10+
+- [`uv`](https://github.com/astral-sh/uv)
+- Linux with cgroups v2 (required for background process management)
```
xaru8145: Not only background process management. We also use it for the script, foreground processes, etc. We can just say "Linux with cgroups v2" without adding the "(required for background process management)". Maybe link to cgroups v2 docu if you haven't done that already anywhere else: https://docs.kernel.org/admin-guide/cgroup-v2.html

README.md
```diff
@@ -11,21 +11,159 @@ Most benchmarking systems are built around a specific algorithm, dataset format,
...
+Run it with the CLI:
+
+```bash
+lambkin my_benchmark.py
+```
+Run `lambkin --help` to see all available options.
+
+For a complete, working example using the Beluga algorithm, see [`examples/beluga/beluga_benchmark.py`](examples/beluga/beluga_benchmark.py). For the full CLI reference, see the [SDK documentation](src/lambkin/README.md#cli).
+
+## Use Cases
```
xaru8145, line 122: We could add the most basic use case, which is parameter sweeps. Then briefly explained that this can be done for programatic tuning of the algorithm parameters, to optimize, to catch regressions, etc.

examples/beluga/README.md
```diff
@@ -38,51 +25,70 @@ The launch file accepts the map path, sensor model type, and particle count as p
...
-### **(Optional) Use your own data**
+### Run in a Container
 
-By default, no volume mounts are needed — the example uses the rosbag, map, and groundtruth baked into the image. To benchmark your own data instead, edit the compose file ([docker-compose.yml](docker/docker-compose.yml) or [podman-compose.yml](docker/podman-compose.yml)) and uncomment the input volume mounts, pointing them at your files:
+If you are running the benchmark as-is, use the **Production** profile. If you are modifying the benchmark script or the ROS 2 package, use the **Development** profile.
 
-Edit `docker/docker-compose.yml` and set the host paths to your reference files:
+#### 1. (Optional) Use your own data
+
+By default, no volume mounts are needed in either runtime — the development image bakes in the rosbag, map, and groundtruth via `COPY --from=ekumenlabs/lambkin-beluga-datasets:jazzy /data/ /data/` in [`docker/Dockerfile`](docker/Dockerfile), and both [`docker-compose.yml`](docker/docker-compose.yml) and [`podman-compose.yml`](docker/podman-compose.yml) build from that same Dockerfile.
```
xaru8145: We do not need to be that specific explaining the command we use to copy the bundled data. The previous approach was simpler and complete, which was:

By default, no volume mounts are needed — the example uses the rosbag, map, and groundtruth baked into the image. To benchmark your own data instead, edit the compose file ([docker-compose.yml](https://github.com/Ekumen-OS/lambkin/blob/next-gen/examples/beluga/docker/docker-compose.yml) or [podman-compose.yml](https://github.com/Ekumen-OS/lambkin/blob/next-gen/examples/beluga/docker/podman-compose.yml)) and uncomment the input volume mounts, pointing them at your files:

examples/beluga/README.md
```diff
@@ -38,51 +25,70 @@ The launch file accepts the map path, sensor model type, and particle count as p
...
+
+> [!NOTE]
+> That `COPY --from=...` line only exists in the `development` stage of the Dockerfile, not `production`. If you rely on the bundled dataset, use the Development profile, or check with the team before assuming it's also available in Production.
```
xaru8145: This note is not true. It is available in both, you fixed it. Please remove the note.

xaru8145 (changes requested):

README.md
teresa-ortega: Solved in 7207f2e

examples/beluga/README.md
teresa-ortega: Solved in 7207f2e

src/lambkin/README.md
teresa-ortega: Solved in d66ea4d

src/lambkin/README.md
teresa-ortega: Sorry, solved in 3363e0c

README.md
teresa-ortega: Solved in d0c8f57

README.md
teresa-ortega: SOlved in c9257f6

README.md
teresa-ortega: Solved in 40e7598

examples/beluga/README.md
teresa-ortega: Solved in 02c9e2f

README.md
teresa-ortega: Solved in 2005f8c

examples/beluga/README.md
teresa-ortega: Solved in 27fc3b6

README.md
teresa-ortega: Solved in 31b728f

README.md
```diff
@@ -11,21 +11,182 @@ Most benchmarking systems are built around a specific algorithm, dataset format,
...
+| **Result access** | Read structured benchmark outputs (`lambkin.data`) from a notebook or a standalone script, with no live benchmark required. |
 
 To understand how LAMBKIN works under the hood, see the [SDK documentation](src/lambkin/README.md).
 
 
+## Prerequisites
+
+- Python 3.10+
+- [`uv`](https://github.com/astral-sh/uv)
+- [Linux with cgroups v2](https://docs.kernel.org/admin-guide/cgroup-v2.html)
```
xaru8145:
```suggestion
- Linux with [cgroups v2](https://docs.kernel.org/admin-guide/cgroup-v2.html)
```

README.md
```diff
@@ -111,6 +121,20 @@ For a complete, working example using the Beluga algorithm, see [`examples/belug
 
 ## Use Cases
 
+### Parameter sweeps
+
+The most basic use case: sweep a single parameter across a range of values, running every value under identical conditions. This is the basis for programmatic tuning — searching for the setting that optimizes a metric — and for regression testing, where re-running a fixed sweep over time surfaces any change that degrades performance.
```
xaru8145: Could be a single one or a set of them. Do not scope it down to just one.

README.md
```diff
@@ -111,6 +121,20 @@ For a complete, working example using the Beluga algorithm, see [`examples/belug
...
+
+```python
+@lambkin.benchmark(
+    variants=lambkin.common.named_product(num_particles=[100, 500, 1000, 5000]),
+    num_iterations=30,
+)
+def nominal(ctx):
+    # run the algorithm configured with ctx.variant.num_particles
```
xaru8145: Mabe a reference to the actual algorithm call, for completeness. Like:

ctx.shell.my_algorithm(f"num_particles:={ctx.variant.num_particles}")
or
ctx.shell.my_algorithm(num_particles=ctx.variant.num_particles)

src/lambkin/README.md
```diff
@@ -2,52 +2,126 @@
...
+        └── my_recorder.stderr.log
+```
+
+`lambkin_metadata.yaml` is always written by the SDK itself. Everything else under `iter_<n>/` is whatever your benchmark function's commands wrote to the current working directory — the exact names and shapes depend entirely on the tools you call (e.g. `ros2 bag record -o output` creates an `output/` *directory* with its own internal files, not a single `output.mcap`). Artifacts written by `@output` hooks are not placed anywhere automatically — it's up to the hook to decide where to write them (e.g. `ctx.base_dir / "plots.png"` to put them under `results/`).
+
+> [!NOTE]
+> Two simultaneous `lambkin` invocations against the same `results/` directory are not guarded against — there is no file locking, so concurrent runs may corrupt each other's output. To run two benchmarks in parallel, place them in separate source directories or rename the existing `results/` folder first.
+
+`results/` being tied to the benchmark script's location is a deliberate design choice: it means the cache always works without requiring the user to pass an explicit output path, and keeps each benchmark script bound to a fixed, reproducible setup.
```
xaru8145, line 52: Move the paragraph 

"`results/` being tied to the benchmark script's location is a deliberate design choice: it means the cache always works without requiring the user to pass an explicit output path, and keeps each benchmark script bound to a fixed, reproducible setup."

above the [!NOTE] please

src/lambkin/README.md
```diff
@@ -2,52 +2,126 @@
...
+        ├── my_algorithm.stdout.log
+        ├── my_algorithm.stderr.log
+        ├── my_recorder.stdout.log
+        └── my_recorder.stderr.log
+```
+
+`lambkin_metadata.yaml` is always written by the SDK itself. Everything else under `iter_<n>/` is whatever your benchmark function's commands wrote to the current working directory — the exact names and shapes depend entirely on the tools you call (e.g. `ros2 bag record -o output` creates an `output/` *directory* with its own internal files, not a single `output.mcap`). Artifacts written by `@output` hooks are not placed anywhere automatically — it's up to the hook to decide where to write them (e.g. `ctx.base_dir / "plots.png"` to put them under `results/`).
+
+> [!NOTE]
+> Two simultaneous `lambkin` invocations against the same `results/` directory are not guarded against — there is no file locking, so concurrent runs may corrupt each other's output. To run two benchmarks in parallel, place them in separate source directories or rename the existing `results/` folder first.
```
xaru8145: Simplify the note — "in parallel" implied a supported, deliberate use case, when it's really about avoiding an accidental collision. Also cut the mechanism explanation (how a second invocation gets fooled into rerunning); the instruction is what matters here, not the internals. Added "so the cache stays trustworthy and outputs don't get corrupted" to keep the why without it.

```suggestion
> [!NOTE]
> Avoid running two benchmarks against the same `results/` directory at once — place them in separate source directories, or rename the existing `results/` folder first, so the cache stays trustworthy and outputs don't get corrupted.
```

xaru8145 (changes requested):
Almost there

xaru8145 (approved):
Approving to unblock you but please remember to address the last review comments before merging.

src/lambkin/README.md
teresa-ortega: Solved in 6917903

src/lambkin/README.md
teresa-ortega: Solved in 6917903

README.md
teresa-ortega: Solved in 6917903

README.md
teresa-ortega: Solved in 6917903

README.md
teresa-ortega: Solved in 6917903

PR #179, opened 2026-06-22
Phase 5: Fix podman run command 

teresa-ortega:
### Proposed changes

Fix `--podman-run-args` flag position in Podman run commands and update the Dockerfile to use fully qualified image names for Podman compatibility.

#### Fix `--podman-run-args` position

The `--podman-run-args` flag must be placed before the subcommand. Updated both Development and Production Podman commands in the README to move `--podman-run-args="--systemd=always"` before `run`.

#### Fix Dockerfile registry prefix

Update `FROM` lines in the Dockerfile to use fully qualified image names with the `docker.io` registry prefix.

Unlike Docker, Podman does not have Docker Hub as an implicit default registry. Running the following command on a system without `unqualified-search-registries` configured in `/etc/containers/registries.conf` results in:

```
podman-compose --profile development --podman-run-args="--systemd=always" run --rm lambkin_dev bash
[1/2] STEP 1/6: FROM ros:jazzy-ros-base AS base
Error: creating build container: short-name "ros:jazzy-ros-base" did not resolve to an alias and no unqualified-search registries are defined in "/etc/containers/registries.conf"
Error: short-name "lambkin_ros_dev:jazzy" did not resolve to an alias and no unqualified-search registries are defined in "/etc/containers/registries.conf"
```

**Options considered:**

1. **Update `FROM` lines in the Dockerfile** to use fully qualified image names (e.g. `FROM docker.io/ros:jazzy-ros-base`). ✅ Chosen
2. **Build the image with Docker and pull it into Podman** via `podman pull docker-daemon:lambkin_ros_dev:jazzy`. Requires Docker to be installed and running.
3. **Add Docker Hub to `unqualified-search-registries`** in `/etc/containers/registries.conf`. Requires system-level configuration on each machine.

Option 1 is the most portable solution — it requires no extra tooling or system configuration, and no documentation changes are needed.

#### Type of change

💥 No breaking changes

### Checklist

### Additional comments

N/A

### DIFF
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -97,7 +97,7 @@ docker compose --profile development exec lambkin_dev bash
 **Podman**
 
 ```bash
-podman-compose --profile development run --podman-run-args="--systemd=always" --rm lambkin_dev bash
+podman-compose --profile development --podman-run-args=--systemd=always run --rm lambkin_dev bash
 ```
 
 > [!NOTE]
@@ -127,7 +127,7 @@ docker compose --profile production run --rm lambkin_prod bash
 **Podman**
 
 ```bash
-podman-compose --profile production run --podman-run-args="--systemd=always" --rm lambkin_prod bash
+podman-compose --profile production --podman-run-args=--systemd=always run --rm lambkin_prod bash
 ```
 
 > [!WARNING]
--- a/examples/beluga/README.md
+++ b/examples/beluga/README.md
@@ -97,7 +97,7 @@ docker compose --profile development exec lambkin_dev bash
 **Podman**
 
 ```bash
-podman-compose --profile development --podman-run-args=--systemd=always run --rm lambkin_dev bash
+podman-compose --profile development --podman-run-args="--systemd=always" run --rm lambkin_dev bash
 ```
 
 > [!NOTE]
@@ -127,7 +127,7 @@ docker compose --profile production run --rm lambkin_prod bash
 **Podman**
 
 ```bash
-podman-compose --profile production --podman-run-args=--systemd=always run --rm lambkin_prod bash
+podman-compose --profile production --podman-run-args="--systemd=always" run --rm lambkin_prod bash
 ```
 
 > [!WARNING]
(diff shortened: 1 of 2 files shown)


xaru8145 (commented):
There are 2 changes here:
- Moving `--podman-run-arg`s before `run`
- Quote removal.

Are both really needed?

Some questions:
- Can you explain how did you realize the previous command was not working? 
- Is there any documentation or changelog that describes the usage without quotes explicitely?

teresa-ortega:
Both changes are needed, but for different reasons.

**Quote removal:** Running the original command produced this error:

```
podman-compose: error: argument --podman-run-args: expected one argument
```

With `--podman-run-args="--systemd=always"`, the shell strips the outer quotes but argparse still sees the value starting with `--`, which it interprets as a new flag rather than the argument to `--podman-run-args`. Using `=` syntax (`--podman-run-args=--systemd=always`) avoids that ambiguity and passes the value correctly.

**Flag position (before `run`):** Running with the quotes fixed but flag after `run` produced:

```
podman-compose: error: unrecognized arguments: --podman-run-args=--systemd=always
```

This is standard argparse behavior: once the subcommand (`run`) is consumed, argparse stops processing global flags. `--podman-run-args` is a global flag (defined at the top-level parser, not the `run` subparser), so it must come before the subcommand.

This behavior is visible in the argparse definition in `podman_compose.py` — `--podman-run-args` is added to the top-level parser alongside `--podman-args`, `--profile`, etc., and those all appear before `{up,down,run,...}` in the usage string podman-compose itself prints on error.

There is no changelog entry or explicit documentation describing this, I discovered both issues empirically from the runtime errors above.

xaru8145:
The release notes act as a changelog. See https://github.com/containers/podman-compose/releases/tag/v1.5.0

I do not see any change in that command so maybe argparse version has changed. Anyway, it is strange as there are some issues that mention the usage of this flag using quotes. For example: https://github.com/containers/podman-compose/issues/799

xaru8145 (approved):

teresa-ortega:
You were right. We tested with the quotes and moving only the flag position, and it works:

```
podman-compose --profile production --podman-run-args="--systemd=always" run --rm lambkin_prod bash
```

So only one change was needed: moving `--podman-run-args` before the subcommand. Removing the quotes was unnecessary. The README has been updated accordingly.

xaru8145:
Perfect! Can you update the PR description accordingly as well? @teresa-ortega

teresa-ortega:
Done, the PR description has been updated.

xaru8145 (approved):

PR #180, opened 2026-06-23
Phase 6: Report generation

teresa-ortega:
### Proposed changes

Adds `examples/beluga/report.ipynb`, a Jupyter cookbook that shows how to explore and visualise APE results from a completed Beluga benchmark run using `lambkin.data.access` and `lambkin.data.evo`.

The notebook covers:

- **Explore available data** — list variants and iterations using `access.iterations()`.
- **APE timeseries by variant** — individual iterations in light color, per-variant mean overlaid in bold.
- **Stats summary table** — RMSE mean/std, mean, and max per variant, aggregated across iterations.
- **RMSE comparison** — bar chart with error bars across variants.
- **Pandas DataFrame** — snippet showing how to convert results to a long-format DataFrame for advanced analysis with seaborn.
- **Export to HTML** — how to share the notebook as a standalone HTML file with `jupyter nbconvert`.

No new SDK modules or dependencies are introduced — the notebook uses only the existing `lambkin.data` API.

#### Type of change

💥 No breaking changes

### Checklist

### Additional comments

N/A


xaru8145:
Before we go further in the review of this implementation, I want you to think through some scenarios:
1. What happens when a user wants RPE instead of APE?
2. What happens when they want both APE and RPE in the same report?
3. What happens when they only care about RMSE and not the full stats table?

Walk through how a user would achieve each of these with the current `report.generate()` API. What does that tell you about the generality of this approach?

teresa-ortega:
You're right, the current `report.generate()` is too opinionated — it hardcodes APE, always generates all three sections, and can't combine multiple metrics in the same report.

I see two ways to fix this:

**Option A : composable cells**

Instead of a single `generate()`, the SDK exposes building blocks that the user composes:

```python
import lambkin.data.report as report

nb = report.notebook()
nb.add(report.timeseries("output.ape.zip", title="APE"))
nb.add(report.timeseries("output.rpe.zip", title="RPE"))
nb.add(report.stats_table("output.ape.zip"))
nb.save(output_dir / "report.ipynb")
```

This handles all three scenarios cleanly. More API surface, but fully flexible.

**Option B : `generate()` with parameters**

Keep a single entry point but make it configurable:

```python
report.generate(
    ctx,
    files=["output.ape.zip", "output.rpe.zip"],
    sections=["timeseries", "stats"],
)
```

Which direction do you prefer?

src/lambkin/README.md
```diff
@@ -303,7 +303,7 @@ Precedence (highest to lowest):
 - **`lambkin.data.access.iterations(source)`** — walks `results/var_*/iter_*/`, skips any iteration that didn't complete, and returns one entry per completed iteration with `iter_dir`, `variant` (e.g. `"var_1"`), `iteration`, and `params` (the variant's parameters as a `SimpleNamespace`). Accepts either a context-like object exposing `.base_dir`, or a plain path/string.
 - **`lambkin.data.evo.series(source, filename)`** — same traversal, plus loads the `evo` result file (e.g. `"output.ape.zip"`) from each iteration directory and exposes `time`, `error`, and `distance` arrays, ready to plot.
 - **`lambkin.data.evo.stats(source, filename)`** — same traversal, but exposes the aggregate statistics `evo` computes for each result: `rmse`, `mean`, `median`, `std`, `min`, `max`, `sse`.
-
+- **`lambkin.data.report.generate(source, filename, output_dir)`** — builds on the two functions above to write a ready-to-run `report.ipynb` notebook summarizing a benchmark run. See [Report Generation](#report-generation) below.
```
xaru8145: This is a leftover from previous approach.

src/lambkin/README.md
```diff
@@ -348,6 +348,27 @@ def stats(ctx):
...
+`lambkin.data.access` and `lambkin.data.evo`.
+
+The cookbook covers:
+
+1. Listing available variants and iterations with `access.iterations()`.
+2. APE timeseries by variant — individual iterations + per-variant mean.
+3. Stats summary table — RMSE, mean, and max aggregated across iterations.
+4. RMSE comparison bar chart across variants.
+5. Converting results to a pandas DataFrame for advanced analysis with seaborn.
+6. Exporting the notebook to HTML with `jupyter nbconvert`.
```
xaru8145: Cookbook vs example — these are different things. An example shows a complete working case ("here's a benchmark that does X"). A cookbook shows reusable patterns ("here's how to plot APE across variants"). The Beluga notebook sounds like it's trying to be both, which makes it hard to know what the reader should take away.

The cleaner split would be: the Beluga benchmark directory contains a `report.ipynb` that is explicitly scoped to Beluga's outputs (a report template, not a cookbook), and separately, the SDK documentation has a cookbook section with short focused snippets showing individual patterns — how to call evo.stats(), how to plot timeseries, how to export. Those don't live in a notebook, they live in the docs.

examples/beluga/report.ipynb
xaru8145, line 1: The notebook has been committed with cell outputs already rendered. Since the intended workflow is to run the benchmark first and then execute the notebook, committed outputs are misleading — they show a specific past run rather than the user's own results, and will go stale. Please strip outputs before committing. The notebook should be a reusable tool, not a snapshot of one execution.

The committed output also leaks a contributor's local filesystem path (`/home/teresa/ekumen/`...), which is a good example of why notebook outputs shouldn't be committed at all — they capture machine-specific state that has no place in the repository.

xaru8145 (changes requested):

src/lambkin/README.md
teresa-ortega: Solved in ddc5baf

src/lambkin/README.md
teresa-ortega: Solved in 02a0432

examples/beluga/report.ipynb
teresa-ortega: Solved in deb17b5

src/lambkin/README.md
```diff
@@ -410,3 +409,46 @@ def nominal(ctx):
 
 > [!NOTE]
 > `ShellProxy` converts keyword argument underscores to dashes (`save_as_tum=` → `--save-as-tum`), which `evo` won't recognize. Always pass `evo` flags that contain underscores as positional strings, as shown above.
+
+### Analysing results in a notebook
+
+`lambkin.data` functions accept a plain path, so results can be explored
+from a Jupyter notebook without re-running the benchmark:
```
xaru8145, line 416: Framing: The subject should be the notebook as a workflow, not lambkin.data. Something like:

> Benchmark results can be explored interactively in a Jupyter notebook after a run. The included report.ipynb provides a starting point; lambkin.data functions accept a plain path so no benchmark context is needed:

src/lambkin/README.md
```diff
@@ -410,3 +409,46 @@ def nominal(ctx):
...
+plt.show()
+
+# Convert to a long-format DataFrame for seaborn
+rows = []
+for entry in evo_data.stats(RESULTS_DIR, "output.ape.zip"):
+    row = vars(entry.params).copy()
+    row.update({"variant": entry.variant, "iteration": entry.iteration,
+                "rmse": entry.rmse, "mean": entry.mean, "max": entry.max})
+    rows.append(row)
+df = pd.DataFrame(rows)
```
xaru8145, line 447: The code snippet also reads like a Python script — using print() and plt.show() — which works against the notebook framing. In a notebook, cell outputs render inline; you don't need either. 

Rather than writing a new snippet, just point to the Beluga example notebook and note that it can be adapted for different metrics. That's faster and more honest about what the user should actually do.

examples/beluga/README.md
```diff
@@ -166,3 +166,13 @@ On top of the [SDK-wide options](../../src/lambkin/README.md#cli), this benchmar
...
+
+### Explore Results
+
+After the benchmark completes, open the report notebook to visualise APE results across all variants and iterations:
+
+```bash
+jupyter notebook examples/beluga/report.ipynb
+```
+
+The notebook plots APE timeseries by variant, prints a stats summary table, and generates an RMSE comparison bar chart. All figures are saved under `results/`.
```
xaru8145, line 178: in a notebook, outputs render inline, nothing is "printed". Change to "displays" or "shows".
```suggestion
The notebook plots APE timeseries by variant, shows a stats summary table, and generates an RMSE comparison bar chart. All figures are saved under `results/`.
```

xaru8145 (changes requested):

PR #187, opened 2026-09-02
Remove the GPL-3 evo library dependency

xFranv8:
### Proposed changes

LAMBKIN is Apache-2.0 but depended on `evo`, which is GPL-3, and imported it in-process. The import
was eager, so every user loaded GPL code even when they never evaluated a trajectory.

The only thing we used from the library was `file_interface.load_res_file`, to read the result `.zip`
that `evo_ape` writes. This replaces it with our own reader, `lambkin.data.results`, and drops `evo`
from the dependencies. The `evo_*` binaries are still run as external processes, unchanged.

`lambkin.data.evo.series()` and `.stats()` keep their signatures and behaviour, so benchmark scripts
don't need changing.

The reader was written from observation of archives the evo tools produce, not from evo's source.
`test/data/fixtures/` holds real `evo_ape` and `evo_rpe` output that the tests read back, so the
reader is checked against genuine archives and not just against our own writer.

Two side effects worth knowing about:

- `numpy` and `pyyaml` are now declared. `pyyaml` was already used in production code and only
  worked because `evo` pulled it in.
- The beluga image installs `evo` and `matplotlib` explicitly, since they used to arrive with `evo`.

#### Type of change

### Checklist

- [x] I have not added a copyleft dependency to `pyproject.toml`

### Additional comments

@agalbachicar


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -18,6 +18,7 @@ _Put an `x` in the boxes that apply. This is simply a reminder of what we will r
 - [ ] Lint and unit tests (if any) pass locally with my changes
 - [ ] I have added tests that prove my fix is effective or that my feature works
 - [ ] I have added necessary documentation (if appropriate)
+- [ ] I have not added a copyleft dependency to `pyproject.toml`
```
agalbachicar, line 21: Nice.

agalbachicar (approved):
LGTM, thanks @xFranv8 
Please check with one of the maintainers @hidmic @glpuga @xaru8145 for an in depth review. I guess it just makes things right.

src/lambkin/data/results.py
```diff
@@ -0,0 +1,169 @@
...
+        path: Destination path, overwritten if it exists.
+        result: The result to serialize.
+
+    Raises:
+        ResultFormatError: If an array cannot be stored without pickling.
+    """
+    path = Path(path)
+    with zipfile.ZipFile(path, "w", zipfile.ZIP_DEFLATED) as archive:
+        archive.writestr(INFO_MEMBER, json.dumps(result.info))
+        archive.writestr(STATS_MEMBER, json.dumps(result.stats))
```
glpuga, line 134: nit: This will use no-indentation, making it hard to read. Maybe add indentation to make the file human-readable.

glpuga (approved):
LGTM

glpuga:
@xFranv8 let me know if this is ready for merging and you don't have permission to hit "Merge".

xFranv8:
> @xFranv8 let me know if this is ready for merging and you don't have permission to hit "Merge".

I don't have permissions to merge, but it is ready.

PR #188, opened 2026-09-03
Package LAMBKIN and evo with Nix

xFranv8:
> **Stacked on #187**, whose branch only exists on the fork — so this PR's base is `main`
> and the diff carries #187's two commits too. Review the last commit, or merge #187 first
> and this will shrink.

### Proposed changes

Makes LAMBKIN installable with Nix, no clone required:

```bash
nix profile install github:Ekumen-OS/lambkin        # Apache-2.0
nix profile install github:Ekumen-OS/lambkin#evo    # GPL-3, opt-in
```

`evo` gets its own derivation because it isn't in nixpkgs, and stays a separate output to
keep #187's licensing split: nothing wraps or aggregates the two, and `evo` never enters
LAMBKIN's closure. Since it is GPL-3, its output ships the licence text and the complete
corresponding source in `share/doc/evo/`. `examples/` is not packaged.

Everything was written and verified inside `nix/docker/`, a pinned `nixos/nix` image
committed alongside the flake, so anyone with Docker can reproduce the build:

```bash
./nix/docker/run.sh build .#lambkin
./nix/docker/run.sh flake check --all-systems
./nix/docker/run.sh smoke
```

The new `nix` workflow runs that same script, with the Nix build sandbox on.

#### Worth a reviewer's attention

- **`nix/lambkin.nix` patches the `uv_build` bound.** nixpkgs ships 0.10.0, `pyproject.toml`
  asks for `>=0.10.3,<0.11.0`, and `pypaBuildHook` enforces it. `pythonRelaxDeps` can't
  help — it rewrites the wheel's `METADATA`, not `[build-system].requires`.
- **`pydantic` is dropped from the Nix closure.** It's declared but imported nowhere under
  `src/`. Removing the declaration itself would touch `uv.lock`, so that's a `TODO`.
- **The CLI goes through a `withPackages` env, not `toPythonApplication`.** `cli.py` re-execs
  the benchmark with `sys.executable`; a plain wrapper only works via an exported
  `PYTHONPATH` that then leaks into `ros2 launch` and `evo_ape`. Verified end to end with a
  real benchmark under `--privileged --cgroupns=private`.
- **evo is pinned to v1.35.2**, which asks for `rosbags>=0.11.0` — exactly what nixpkgs has.
  v1.36+ raised that floor, and relaxing it would turn a build error into a mid-benchmark
  `AttributeError`.
- **`nix flake check` runs the cgroup-free tests** (56, from `test/common`, `test/utils`,
  `test/data`). The rest needs a writable cgroup v2 hierarchy and stays with the podman
  `pytest` workflow. No `ruff`/`mypy` checks — pre-commit already runs both with exact pins.

#### Type of change

### Checklist

- [x] I have not added a copyleft dependency to `pyproject.toml`

### Additional comments

`flake.lock` pins `nixos-26.05`. `--rebuild` confirms both derivations are bit-reproducible.

@agalbachicar


flake.nix
```diff
@@ -0,0 +1,86 @@
...
+        lambkin = {
+          type = "app";
+          program = lib.getExe pkgs.lambkin;
+          inherit (pkgs.lambkin) meta;
+        };
+        evo = {
+          type = "app";
+          program = lib.getExe pkgs.evo;
+          inherit (pkgs.evo) meta;
+        };
```
agalbachicar, line 58: For my own understanding, why is it evo here?

README.md
```diff
@@ -69,6 +72,31 @@ uv sync
...
+Alternatively, with [Nix](https://nixos.org) — no clone required:
+
+```bash
+nix profile install github:Ekumen-OS/lambkin
+nix run github:Ekumen-OS/lambkin -- my_benchmark.py
+```
+
+`evo` is a separate output, so LAMBKIN's Apache-2.0 closure never contains GPL-3 code. A
+fresh install has no `evo_*` on `PATH`, and a benchmark that calls one fails with `Command
+not found`, so install it when you need it:
```
agalbachicar:
```suggestion
`evo` is a separate output. A fresh install has no `evo_*` on `PATH` thus a benchmark that calls one
fails with `Command not found`.
Install it when you need it:
```

nix/evo.nix
agalbachicar, line 1: Strictly speaking this might need to leave somewhere else, as this is not a packaging configuration for lambkin but evo and this will trigger a GPL 3.0 on this file and the output bundle. Maybe move to its own repository?  I can provide the bits for it.

nix/overlay.nix
```diff
@@ -0,0 +1,32 @@
...
+    (python-final: _python-prev: {
+      lambkin = python-final.callPackage ./lambkin.nix { };
+      evo = python-final.callPackage ./evo.nix { };
+    })
+  ];
+
+  # Separate top-level attributes on purpose: LAMBKIN only invokes evo_* as
+  # external processes on PATH, so evo's GPL-3 closure never meets LAMBKIN's.
+  lambkin = final.callPackage ./lambkin-cli.nix { };
+  evo = final.python3.pkgs.toPythonApplication final.python3.pkgs.evo;
```
agalbachicar, line 31: Can we make this user benchmark specific? It can be another package on its own that brings evo and lambkin as dependencies and treats the evo recipe as a dependency in the end.

nix/lambkin.nix
```diff
@@ -0,0 +1,91 @@
...
+      ../LICENSE
+      # Filtering by extension drops __pycache__ without risking
+      # src/lambkin/core/shell/ros/, which has no __init__.py.
+      (fs.fileFilter (file: file.hasExt "py" || file.hasExt "md") ../src)
+    ];
+  };
+in
+buildPythonPackage {
+  pname = "lambkin";
+  version = "0.0.0";
```
agalbachicar: This is a problem. We should release a v0.0.2 maybe?

agalbachicar (changes requested):
PTAL and tell me what you think @xFranv8

flake.nix
xFranv8: It's what makes nix run github:Ekumen-OS/lambkin#evo work, it runs the bare evo command,

nix/evo.nix
xFranv8: Happy to move it out, send me the repo and I'll delete this file and pull evo in as a flake input instead.

README.md
xFranv8: Done in https://github.com/Ekumen-OS/lambkin/pull/188/commits/1eafc5f9a7bcc39cd1fbb6e09132f4461a842e52

nix/overlay.nix
xFranv8: I'd do this in the same follow-up that adds the evo input, since right now there's nothing external for the benchmark package to depend on.

nix/lambkin.nix
xFranv8: Done in https://github.com/Ekumen-OS/lambkin/pull/188/commits/1eafc5f9a7bcc39cd1fbb6e09132f4461a842e52

